PR #87, opened 2014-08-06
Add support for checksumming network sources

### LINKED ISSUE #21: Accept hashes for URIs
`URI` objects should allow (and strongly encourage) specifying a sha1 hash that is verified when the associated file is downloaded.

staticfloat:
Closes #21 
For example usage, see https://github.com/staticfloat/Nettle.jl/commit/b0a3449d48ba0b1b6af37d2c0d1dd7f6ec270087


tkelman:
This needs documentation. At least an example. And somewhere either here or in SHA.jl is keeping the file open, which means we have to restart Julia to delete it.

staticfloat:
@tkelman Usage is simple:

Add a `SHA` field to your `Sources` declaration, e.g. change:
```
provides( Sources,
          URI("http://www.lysator.liu.se/~nisse/archive/nettle-2.7.1.tar.gz"),
          nettle )
```
to:
```
provides( Sources,
          URI("http://www.lysator.liu.se/~nisse/archive/nettle-2.7.1.tar.gz"),
          SHA="bc71ebd43435537d767799e414fce88e521b7278d48c860651216e1fc6555b40",
          nettle )
```

If there is no `SHA` field provided, checksumming will be skipped.

PR #107, opened 2014-10-08
Track Julia 0.4 Dict changes

garborg:
Get rid of the warnings on 0.4 without affecting 0.3. Easier to look at just the second commit, if you're interested (the first one caught a lot of tabs and trailing whitespace).


musm:
This is probably no longer relevant?

garborg:
Indeed, thanks for the bump.

PR #115, opened 2014-11-14
add support for the Nix package manager

rened:
Nix is a package manager which can be used on various UNIXes, it is also the basis of the NixOS distribution.

One of its strengths is the fact that non-root users can install packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ defined the following dependencies:
 ```
 
 Let's suppose that these libraries are available in the `libfoo-dev` and `libbaz-dev`
-in apt-get and that both libraries are installed by the `libbaz` yum package and the `baz` pacman package. We may
+in apt-get and that both libraries are installed by the `libbaz` yum package and the `baz` pacman or nix package. We may
 declare this as follows:
 
 ```julia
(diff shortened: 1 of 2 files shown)


mk:
hey there! We're working on offering a hosted notebook environment for julia using non root containers and would be very interested in getting this to work. Anything we can do to help?

tbenst:
Thought I'd bump this. Most of Julia works flawlessly on NixOS, but I just tried to install HDF5.jl which has BinDeps as a dependency, and the install failed due to the use of `ldconfig -p`, which is not used on NixOS. I tried naively just commenting this code out, and the code fails at the following step: `/home/tyler/.julia/packages/HDF5/H0XJB/deps/src/hdf5-1.10.0-patch1/configure --prefix=/home/tyler/.julia/packages/HDF5/H0XJB/deps/usr`, which I'm guessing is due to a missing dependency but it's hard to tell without seeing the subcommand output.

StefanKarpinski:
This PR is pretty out of date at this point and it seems like the change needs to be with [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl) or [BinaryProvider](https://github.com/JuliaPackaging/BinaryProvider.jl) not with Julia itself, so I'm not sure there's anything to do here anymore, is there?

thomasjm:
FWIW some packages still use BinDeps. For example, Cairo.jl still does (they've had a [PR](https://github.com/JuliaGraphics/Cairo.jl/pull/229) to migrate to BinaryProvider for almost a year).

@tbenst 's solution is very helpful. But maybe this PR should still be landed since BinDeps may not go away anytime soon.

PR #124, opened 2015-01-20
Add Powershell option for download_cmd

### LINKED ISSUE #121: Don't use curl to download files on Windows.
BinDeps currently uses the curl binary that Julia bundles along with command-line git to download some files. This causes problems for people operating behind a proxy (see JuliaLang/julia#3001), and Julia eventually plans to stop distributing curl. For a potential alternative, see the Windows APIs that WinRPM.jl uses to download files. @tkelman has further background on this issue.

tkelman:
Should fix #121 

@djpeaco could you give this a try? Do `Pkg.checkout("BinDeps", "tk/powershell")`, restart Julia, then try to do `Pkg.build` on any packages that had been causing trouble with your proxy.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -45,7 +45,7 @@ module BinDeps
     function download_cmd(url::String, filename::String)
         global downloadcmd
         if downloadcmd === nothing
-            for checkcmd in (:curl, :wget, :fetch)
+            for checkcmd in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 try
                     if success(`$checkcmd --help`)
                         downloadcmd = checkcmd
@@ -62,6 +62,8 @@ module BinDeps
             return `curl -o $filename -L $url`
         elseif downloadcmd == :fetch
             return `fetch -f $filename $url`
+        elseif downloadcmd == :powershell
+            return `powershell -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
         else
             error("No download agent available; install curl, wget, or fetch.")
         end


djpeaco:
Sorry for the extreme delay in testing this. The following works:
> Pkg.update()
> Pkg.add("BinDeps")
> Pkg.checkout("BinDeps")
> Pkg.checkout("BinDeps", "tk/powershell")
> Pkg.add("IJulia")
> Pkg.build("IJulia")
> using IJulia
> notebook()

tkelman:
No big rush I guess, thanks for testing. Was this failing due to your proxy if you didn't do `Pkg.checkout("BinDeps", "tk/powershell")` ?

tkelman:
Note that we did change the ZMQ binary to use WinRPM, so that may not be problematic any more. Blosc.jl I think is using an mit.edu address for its binary download.

djpeaco:
Ah, that's a good point. I tried `Pkg.add("Blosc"); Pkg.test("Blosc")`. It failed to download/install using stock BinDeps, but worked using your branch (after I remembered I needed to restart julia).

tkelman:
Awesome, thanks!

staticfloat:
This is great, thanks guys!

tkelman:
One slightly funny thing this seems to do is change the process title of Julia to "Windows PowerShell." I wonder how to make it not do that...

staticfloat:
Unfortunately, I don't think we can use terminal escape codes to set the title here. :P

tkelman:
I think @ihnorton implemented the function to change the process title in base recently, but it's only on master.

ihnorton:
What was the reason to use powershell here instead of the built-in `download` command?

tkelman:
Less refactoring to use something that's shelling out to a subprocess, considering the way the rest of the code works. `Base.download` does not follow redirects, last I checked.

PR #131, opened 2015-03-16
Only accept a single library for each name, and move SystemPath out of Binaries

staticfloat:
Without this check in place, provider preferences were actually quite broken; the least preferred would get picked, and hooks would get mixed with library paths from other providers.  See https://github.com/staticfloat/homebrew-juliadeps/issues/38 for an example.

Moving SystemPath out of Binaries is necessary so that it is not included in `defaults` more than once, and so it has the proper ordering in preference

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -210,7 +210,7 @@ abstract Binaries <: DependencyProvider
 # indicates the library was found somewhere on the
 # system using dlopen.
 #
-immutable SystemPaths <: Binaries; end
+immutable SystemPaths <: DependencyProvider; end
 
 show(io::IO, ::SystemPaths) = print(io,"System Paths")
 
@@ -793,18 +793,22 @@ macro install (_libmaps...)
                                 if !isempty(libs)
                                     for dep in d.deps
                                         !BinDeps.applicable(dep) && continue
-                                        load_cache[dep.name] = libs[dep][2]
-                                        opts = libs[dep][1][2]
-                                        haskey(opts, :preload) && push!(pre_hooks,opts[:preload])
-                                        haskey(opts, :onload) && push!(load_hooks,opts[:onload])
+                                        if !haskey(load_cache, dep.name)
+                                            load_cache[dep.name] = libs[dep][2]
+                                            opts = libs[dep][1][2]
+                                            haskey(opts, :preload) && push!(pre_hooks,opts[:preload])
+                                            haskey(opts, :onload) && push!(load_hooks,opts[:onload])
+                                        end
                                     end
                                 end
                             else
                                 for (k,v) in libs
-                                    load_cache[d.name] = v
-                                    opts = k[2]
-                                    haskey(opts, :preload) && push!(pre_hooks,opts[:preload])
-                                    haskey(opts, :onload) && push!(load_hooks,opts[:onload])
+                                    if !haskey(load_cache, d.name)
+                                        load_cache[d.name] = v
+                                        opts = k[2]
+                                        haskey(opts, :preload) && push!(pre_hooks,opts[:preload])
+                                        haskey(opts, :onload) && push!(load_hooks,opts[:onload])
+                                    end
                                 end
                             end
                         end


staticfloat:
Note that this fixes things on my end, but I haven't tested it fully, especially with regards to the `LibraryGroup` side of things.  It _looks_ okay to me, but historically that's not a great indicator of success with these kind of things.  I'd especially like a look-over from @Keno to make sure that this is actually how it's intended to work.  If you need a refresher on why I'm doing this, check out [this comment](https://github.com/staticfloat/homebrew-juliadeps/issues/38#issuecomment-81171988) where you can see an example of onload hooks leaking from one provider to another.

src/dependencies.jl
```diff
@@ -793,18 +793,22 @@ macro install (_libmaps...)
                                 if !isempty(libs)
                                     for dep in d.deps
                                         !BinDeps.applicable(dep) && continue
-                                        load_cache[dep.name] = libs[dep][2]
-                                        opts = libs[dep][1][2]
-                                        haskey(opts, :preload) && push!(pre_hooks,opts[:preload])
-                                        haskey(opts, :onload) && push!(load_hooks,opts[:onload])
+                                        if !haskey(load_cache, d.name)
```
tkelman: `d.name` or `dep.name`?

staticfloat: Ah, yes.  Good catch.

tkelman:
is the sorting order of `libs` or `d.deps` important here?

staticfloat:
Yes.  The elements closer to the beginning of the `libs` array are preferred.  Without this change, the least preferred paths were getting chosen when multiple possibilities were returned from `_find_library()`.  I believe this change was introduced while @Keno was implementing library groups where you can have multiple separate libraries being dealt with.

tkelman:
If they're dicts, are we sure the insertion order has anything to do with the iteration order? Or are they tuples or something else?

staticfloat:
[`libs` is an `Any[]`](https://github.com/JuliaLang/BinDeps.jl/blob/sf/therecanbeonlyone/src/dependencies.jl#L445).

tkelman:
@Keno and/or @staticfloat, please comment. Is this okay to merge as-is, or should I tag BinDeps for the Libdl fix without this?

staticfloat:
Tag without this. I'd rather wait for confirmation that this is what we
actually want.

tkelman:
Great, thanks, will do.

rsrock:
Bump.  I need the latest BinDeps, but I'd like to have this too.

staticfloat:
@rsrock does that mean that you've tried out this branch?  It worked for all of my use cases, but haven't heard from @Keno.  If you've tried it out and it didn't break anything, and @Keno doesn't comment before the weekend, I'll just merge it and we'll see how things go.

rsrock:
Yeah, it turns out that I've been on this branch since you've made it.  Haven't noticed any issues with it.

On Apr 21, 2015, at 7:28 PM, Elliot Saba <notifications@github.com<mailto:notifications@github.com>> wrote:


@rsrock<https://github.com/rsrock> does that mean that you've tried out this branch? It worked for all of my use cases, but haven't heard from @Keno<https://github.com/Keno>. If you've tried it out and it didn't break anything, and @Keno<https://github.com/Keno> doesn't comment before the weekend, I'll just merge it and we'll see how things go.

—
Reply to this email directly or view it on GitHub<https://github.com/JuliaLang/BinDeps.jl/pull/131#issuecomment-94982875>.

-------------------------------------------------------------------
Ronald S. Rock, Jr., Associate Professor
Dept. of Biochemistry and Molecular Biology, The University of Chicago
GCIS W240, 929 E. 57th St.
Chicago, IL 60637   USA
+1 773.702.0716 (w), +1 773.702.0439 (f)
rrock@uchicago.edu<mailto:rrock@uchicago.edu>

staticfloat:
Merge ahoy.

rsrock:
Thanks!

PR #132, opened 2015-03-16
Fix for new Libdl module

tkelman:
Also fix race condition in .gz extraction (6b69050cca9565d30f3a5d5f6d56e247b0022b36)

Please test this out, it works for me locally on both 0.3 and 0.4 (lots of deprecation warnings but those aren't time critical), but I am not aware of any packages using .gz archives to test the second commit here.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -6,7 +6,13 @@ module BinDeps
             ChangeDirectory, FileDownloader, FileUnpacker, prepare_src,
             autotools_install, CreateDirectory, MakeTargets, SystemLibInstall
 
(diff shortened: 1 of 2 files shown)


IainNZ:
If you want to check out the death-toll, results are up now: http://pkg.julialang.org/pulse.html

staticfloat:
Now I see why the questions in 6b69050.  Feel free to remove the `-k` option, I left it in because I thought it was useful, but the other options don't use it, so it can go.

tkelman:
I don't want to leave this broken too long. With a slight risk of breaking 0.3, I guess the best way to get this tested is to merge and tag. @staticfloat do you want to merge your other fix as well before tagging?

PR #134, opened 2015-03-22
Remove deprecation warnings from BinDeps

jakebolewski:
Depends on https://github.com/staticfloat/SHA.jl/pull/3 and https://github.com/JuliaWeb/URIParser.jl/pull/18 being merged and those packages being tagged first.


jakebolewski:
Hmm, it does not look like this is going to work.  The way macros are eval'd in BinDeps does not play well with Compat.

PR #136, opened 2015-03-24
Remove deprecation warnings from BinDeps

jakebolewski:
This works by not trying to use renamed types in `Compat` (ex. String -> AbstractString).


src/BinDeps.jl
```diff
@@ -1,37 +1,32 @@
...
+
     export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
             Choice, Choices, CCompile, FileDownloader, FileRule,
             ChangeDirectory, FileDownloader, FileUnpacker, prepare_src,
             autotools_install, CreateDirectory, MakeTargets, SystemLibInstall
 
-    if VERSION >= v"0.4.0-dev+3844"
-        import Base.Libdl: dlext, dlpath, RTLD_LAZY, DL_LOAD_PATH
-    else
-        const dlext = isdefined(Base.Sys, :shlib_ext) ? Base.Sys.shlib_ext : Base.Sys.dlext # Julia 0.2/0.3 compatibility
```
tkelman: Unless you test this on 0.2, I think this may need to bump the Julia version in REQUIRE to 0.3

tkelman: or maybe move this line of code to [Compat.jl](https://github.com/JuliaLang/Compat.jl/blob/54ab44412261b7e1de9d0b47242790ac27b13d9e/src/Compat.jl#L280), since 0.2's `Base.Sys` did not have a `dlext`

stevengj: As I noted in #137, support for 0.2 was already broken because `Sys.dlpath` does not exist in 0.2.   So, this is not a regression, and I don't think it should hold up the patch.

(See also JuliaLang/Compat.jl#60)

src/BinDeps.jl
```diff
@@ -1,37 +1,32 @@
...
-    const shlib_ext = dlext # compatibility with older packages (e.g. ZMQ)
+    const dlext = @compat Libdl.dlext
+    const shlib_ext = @compat Libdl.dlext # compatibility with older packages (e.g. ZMQ)
 
     function find_library(pkg,libname,files)
         Base.warn_once("BinDeps.find_library is deprecated, use Base.find_library instead.")
         dl = C_NULL
         for filename in files
-            dl = dlopen_e(joinpath(Pkg.dir(),pkg,"deps","usr","lib",filename))
+            dl = @compat Libdl.dlopen_e(joinpath(Pkg.dir(),pkg,"deps","usr","lib",filename))
```
stevengj: I don't understand why you need `@compat` here.  The Compat package defines a `Libdl` module.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.2-
-URIParser
-SHA
\ No newline at end of file
+URIParser 0.0.5
```
stevengj: As I commented in #137, I don't think we should use the REQUIRE file just to eliminate warnings.  We should only list a version number if that version is required for BinDeps to function.

src/BinDeps.jl
```diff
@@ -182,7 +193,7 @@ module BinDeps
         description::String
         step::SynchronousStepCollection
         Choice(name,description,step) = (s=SynchronousStepCollection();lower(step,s);new(name,description,s))
-    end
```
stevengj: It would make the patch much easier to read if you put all of your whitespace deletions in a separate PR.

src/BinDeps.jl
stevengj: Actually, `dlpath` was not implemented in any form until 0.3, it looks like.  So supporting 0.2 is hopeless (the `dlpath` function uses a C implementation, `jl_pathname_for_handle`, that was not added until 0.3).

REQUIRE
tkelman: agreed here

src/BinDeps.jl
tkelman: try this github trick: https://github.com/JuliaLang/BinDeps.jl/pull/136/files?w=1

jiahao:
Superseded by #137

PR #137, opened 2015-03-26
more 0.4 compatibility fixes, add Compat dependency

stevengj:
This fixes some more 0.4 warnings, e.g. deprecation of `dlopen_e`.  It seemed simplest to just bite the bullet and require the `Compat` package.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
-julia 0.2-
+julia 0.3
 URIParser
 SHA
+Compat 0.3.6
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -1,37 +1,33 @@
 module BinDeps
     importall Base
-
+    using Compat
+    
     export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
             Choice, Choices, CCompile, FileDownloader, FileRule,
             ChangeDirectory, FileDownloader, FileUnpacker, prepare_src,
             autotools_install, CreateDirectory, MakeTargets, SystemLibInstall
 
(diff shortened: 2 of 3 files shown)


REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.2-
 URIParser
-SHA
\ No newline at end of file
+SHA
+Compat
```
jakebolewski: You need to add the lower bound compat version
Compat 0.3.6

It would also be good to add the min versions of URIParser and SHA to get rid of dep warnings in these libraries
URIParser 0.0.5
SHA 0.0.4

jakebolewski:
I already did this, but @tkelman points out that Compat needs
```julia

REQUIRE
stevengj: I don't think we should bound the other version numbers just to get rid of warnings.  The version bounds should only be for things that are *required* to function.  I will fix the Compat version.

stevengj:
Regarding 0.2 compatibility, I also pointed this out in JuliaLang/Compat.jl@8759ca58dd7ed976a6934b4af5d5459dcc2f9b01

However, it is not a regression, because BinDeps had *already* broken 0.2 compatibility by calling `Sys.dlpath`, which doesn't exist either in 0.2.  So, I don't think that 0.2 compatibility should hold up this patch.

stevengj:
Actually, `dlpath` didn't even exist in Julia 0.2; it was added in JuliaLang/julia@d14356eb43bbff7bd85aee9609b6310fc8f54a23

So, we might as well update the REQUIRE to Julia 0.3.

PR #139, opened 2015-04-09
use -f option to curl

porterjamesj:
@jiahao I believe this is ultimately responsible for what happened in https://github.com/porterjamesj/Gumbo.jl/issues/13. Passing `-f`causes 22 to be returned rather than 0 in the case of 4xx or 5xx HTTP response codes, which seems more like what we want here.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -60,7 +60,7 @@ module BinDeps
         if downloadcmd == :wget
             return `wget -O $filename $url`
         elseif downloadcmd == :curl
-            return `curl -o $filename -L $url`
+            return `curl -f -o $filename -L $url`
         elseif downloadcmd == :fetch
             return `fetch -f $filename $url`
         elseif downloadcmd == :powershell


jiahao:
Not silently downloaded a 404 page is an improvement. Does `curl -O -f` save a zero byte file, and does that confuse BinDeps? We might have to test for that explicitly.

porterjamesj:
On my system at least it doesn't. The manual seems to imply that this should be the case in general:

```
       -f, --fail
              (HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal  with
              failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which
              often also describes why and more). This flag will prevent curl from outputting that and return error 22.

              This method is not fail-safe and there are occasions where non-successful response codes will  slip  through,  especially  when
              authentication is involved (response codes 401 and 407).
```

jiahao:
Ok, let's give this a try.

tkelman:
Worth checking what the oldest version of curl is that supports this flag. Hopefully it's universal in commonly used distros.

PR #143, opened 2015-04-25
silence another 0.4 warning

stevengj:
Silences `Warning: both Libdl and Base export dlopen_e; uses of it in module XXXX must be qualified` warnings in recent 0.4 master.


tkelman:
lgtm

StefanKarpinski:
Should the warning maybe not happen when all imports refer to the same object? I.e. When they are === to each other?

PR #147, opened 2015-06-06
Allow providing multiple package names as vector in provides()

nalimilan:
This is often needed to support several distributions.

I needed this this morning for GnuTLS.jl, and I realized it was missing from the docs. ~~You should know whether this actually does work when the line declares several libraries (here`[foo,baz]`), but I haven't actually tested this case.~~

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -298,6 +298,12 @@ provider(::Type{Autotools},a::Autotools; opts...) = a
 provides(provider::DependencyProvider,dep::LibraryDependency; opts...) = push!(dep.providers,(provider,(Symbol=>Any)[k=>v for (k,v) in opts]))
 provides(helper::DependencyHelper,dep::LibraryDependency; opts...) = push!(dep.helpers,(helper,(Symbol=>Any)[k=>v for (k,v) in opts]))
 provides{T}(::Type{T},p,dep::LibraryDependency; opts...) = provides(provider(T,p; opts...),dep; opts...)
+function provides{T}(::Type{T},packages::AbstractArray,dep::LibraryDependency; opts...)
+    for p in packages
+        provides(T,p,dep; opts...)
+    end
+end
+
 function provides{T}(::Type{T},ps,deps::Vector{LibraryDependency}; opts...)
     p = provider(T,ps; opts...)
     for dep in deps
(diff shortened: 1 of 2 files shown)


nalimilan:
Hm, actually that syntax is only supported by the `WinRPM.RPM` provider. Do you agree it would be useful to support it everywhere instead of writing things like `provides(AptGet, Dict("libgnutls28"=>gnutls, "libgnutls-deb0-28"=>gnutls))`? I've added a fix in this PR.

This may well also fix issues with `WinRPM.RPM`. Indeed, it accepts any object for its `package` field, despite the singular (https://github.com/JuliaLang/WinRPM.jl/blob/d376c613c14f513f66338104919f7a95b8e74794/src/bindeps.jl#L9), which could create confusion and probably doesn't work for all functions. With my change, `package` could be restricted to be a single string, and things would still work fine.

src/dependencies.jl
```diff
@@ -269,6 +269,12 @@ provider(::Type{Autotools},a::Autotools; opts...) = a
 provides(provider::DependencyProvider,dep::LibraryDependency; opts...) = push!(dep.providers,(provider,(Symbol=>Any)[k=>v for (k,v) in opts]))
 provides(helper::DependencyHelper,dep::LibraryDependency; opts...) = push!(dep.helpers,(helper,(Symbol=>Any)[k=>v for (k,v) in opts]))
 provides{T}(::Type{T},p,dep::LibraryDependency; opts...) = provides(provider(T,p; opts...),dep; opts...)
+function provides{T}(::Type{T},packages::AbstractArray,dep::LibraryDependency; opts...)
+	for p in packages
```
tkelman: can you change this to use spaces instead of tabs?

nalimilan:
Woops, I had totally forgotten about this. I've removed the tabs, and the mention about OpenSuSE using Yum since we now have a zypper provider. I'll merge soon if nobody objects.

tkelman:
Yeah, I think this makes sense.

nalimilan:
Thanks!

tkelman:
@nalimilan it looks like this broke WinRPM, causing https://github.com/JuliaLang/Cairo.jl/issues/132 - if you don't have any prompt suggestions for how to fix this, I'm likely going to need to revert this for now.

PR #108, opened 2014-10-09
Steps towards Travis testing

timholy:
This would need to have the switch thrown on JuliaLang Travis for this repository, and I don't have rights to do that. CC @StefanKarpinski. (Hopefully there are others reading this who can do that, too.)

That said, currently this is broken. On my machine, `Pkg.test("BinDeps")` yields this output:

``` julia
julia> Pkg.test("BinDeps")
INFO: Testing BinDeps
Morsel is listening on 4653...
INFO: Attempting to Create directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/downloads
INFO: Directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/downloads already created
INFO: Downloading file http://localhost:4653/liba.tar
INFO: Done downloading file http://localhost:4653/liba.tar
INFO: Attempting to Create directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/src
INFO: Directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/src already created
INFO: Attempting to Create directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps
INFO: Directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps already created
INFO: Attempting to Create directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/src/liba
INFO: Directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/src/liba already created
INFO: Attempting to Create directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/builds/liba
INFO: Directory /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/builds/liba already created
INFO: Changing Directory to /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/builds/liba
INFO: Changing Directory to /home/tim/.julia/v0.3/BinDeps/test/testscripts/deps/builds/liba
make: Nothing to be done for `all'.
[...]

PR #149, opened 2015-06-09
Silence deprecation warning: beginswith -> startswith

c42f:
`beginswith()` is deprecated in 0.4 - this causes warnings when building Cairo.jl, for example.  Compat.jl contains a `startswith` definition for 0.3 compatibility.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -113,7 +113,7 @@ function available_versions(p::AptGet)
     vers = ASCIIString[]
     lookfor_version = false
     for l in eachline(`apt-cache showpkg $(p.package)`)
-        if beginswith(l,"Version:")
+        if startswith(l,"Version:")
             try
                 vs = l[(1+length("Version: ")):end]
                 push!(vers, vs)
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks, but don't you need to add `using Compat` at the top of the file?

c42f:
`BinDeps.jl` has `using Compat` at the top, and subsequently does an `include("dependencies.jl")` - `dependencies.jl` isn't designed to be used on its own, so this should be enough, right?

(admittedly, I only tested under 0.4 since I don't have a handy 0.3 install)

nalimilan:
Yeah, I guess you're right.

tkelman:
> (admittedly, I only tested under 0.4 since I don't have a handy 0.3 install)

Please do so as soon as you get a chance.

c42f:
What is the accepted testing method for BinDeps?

timholy:
We should get #108 figured out. Also see #138. We need to merge one of these before we get more of them.

tkelman:
At the moment it has to be manual and indirect. Find a package where running `Pkg.build` with current master of BinDeps goes through these code paths that give deprecation warnings on 0.4. Then run `Pkg.build` (if necessary, clear out `deps/usr` in that package first) from 0.3 on this branch and make sure things still work correctly.

c42f:
Ack, thanks @timholy, It looks like I've just duplicated the patch in #138.  I should have searched the issue list first.

PR #152, opened 2015-06-29
Dead simple tests

IainNZ:
I'm a bit of an interloper in BinDeps land, but I wanted to tag a new version with the fixes for recent deprecations but didn't feel confident doing so without tests. So here is a dead simple setup that just tries to install two packages that probably won't themselves break unless BinDeps itself is broken.

I see that https://github.com/JuliaLang/BinDeps.jl/pull/108 reached a similar conclusion, and I think it'd be great to do finer-grained testing, but I figured that this could be merged right away to get the ball rolling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: julia
+os:
+  - linux
+julia:
+  - release
+  - nightly
+notifications:
+  email: false
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 BinDeps.jl
 ==========
 
+[![Build Status](https://travis-ci.org/JuliaLang/BinDeps.jl.svg?branch=master)](https://travis-ci.org/JuliaLang/BinDeps.jl)
+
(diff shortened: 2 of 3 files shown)


IainNZ:
Oh and I also threw in a REQUIRE bump - I don't think its appropriate to pretend we know if anything works on 0.2 anymore.

timholy:
:+1: from me, I ran out of time to set it up.

IainNZ:
Apparently I lied about the dead simplicity, its failed on the "push" but not the "pr", why would that be?

tkelman:
the parent commit of your branch is nearly a year old, 12dcfab6bddd34e055af9109f311dc1ceec93afc

you must've had a very outdated master that you did `git checkout -b` from

IainNZ:
Haha, wow. OK, fixed now. That fixed the REQUIRE thing too (that was already fixed)

IainNZ:
Will merge when green

tkelman:
sgtm

PR #153, opened 2015-07-01
change DirectoryRule to FileRule in FileUnpacker

kbarbary:
I think this should be a `FileRule` since `joinpath(s.dest, target)` is a file that should exist once the archive is unpacked.


tkelman:
Good catch. I wonder why no one has hit this before? @Keno @staticfloat does this look right? I'll test out this change locally and see if it makes https://github.com/JuliaAstro/FITSIO.jl/issues/42 work, that would be a good sign.

staticfloat:
Yeah, I think that is in fact correct.  Glad to see that it looks like this fixed the issue in `FITSIO.jl`.

Issue #157, opened 2015-07-06
Broke 0.3 and 0.4...


IainNZ:
Seeing some BinDeps breakages on 0.3, but I don't understand the error:

http://pkg.julialang.org/pulse.html

Seeing, e.g.

```
=================================[ ERROR: Cbc ]=================================

File String["/home/vagrant/.julia/v0.3/Cbc/deps/src/Cbc-2.9.1"] was not created successfully (Tried to run SynchronousStepCollection({`tar xzf /home/vagrant/.julia/v0.3/Cbc/deps/downloads/Cbc-2.9.1.tgz --directory=/home/vagrant/.julia/v0.3/Cbc/deps/src`},"","") )
while loading /home/vagrant/.julia/v0.3/Cbc/deps/build.jl, in expression starting on line 62
```

and

```
===============================[ ERROR: LevelDB ]===============================

File String["/home/vagrant/.julia/v0.3/LevelDB/deps/src/leveldb-1.18"] was not created successfully (Tried to run SynchronousStepCollection({`tar xzf /home/vagrant/.julia/v0.3/LevelDB/deps/downloads/v1.18.tar.gz --directory=/home/vagrant/.julia/v0.3/LevelDB/deps/src`},"","") )
while loading /home/vagrant/.julia/v0.3/LevelDB/deps/build.jl, in expression starting on line 30
```

IainNZ:
@tkelman I can't help but feel this is related to the double-build problem, except that this time there was no issues with the VMs and it absolutely should have been clean. Plenty of build-from-source packages passed, for example.

tkelman:
Damn it BinDeps. If these had been passing before, then https://github.com/JuliaLang/BinDeps.jl/issues/153 may have caused a regression. Error looks the same as https://github.com/JuliaLang/BinDeps.jl/issues/155, though from what Simon said that seemingly was no different between the last tag and master of BinDeps?

tkelman:
Okay, reverting https://github.com/JuliaLang/BinDeps.jl/pull/153 seems to fix things locally. So maybe we need to find a different workaround for the FITSIO problem? cc @kbarbary

kbarbary:
Yeah #153 is a problem, now that I look at more of the BinDeps code. The issue is that #153 made the behavior of `FileUnpacker` consistent with the documentation in the comments:
```julia
type FileUnpacker <: BuildStep
    src::String     #file
    dest::String    #directory
[...]

PR #159, opened 2015-07-06
add PathRule for use in FileUnpacker

kbarbary:
This adds a new type `PathRule` for the benefit of `FileUnpacker`. When unpacking an archive, we specify a target that should be created after the file is unpacked. Before #153, that target had to be a directory. For archives that don't contain any directories, we would like to specify a target *file* that should be created. Using `PathRule` instead of `DirectoryRule` or `FileRule` makes it so that the target can be either: it's existence is tested with `ispath` rather than `isdir` or `isfile`.

Hopefully this will fix regressions caused by #153, such as #157.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -140,7 +140,8 @@ module BinDeps
     type FileUnpacker <: BuildStep
         src::String     #file
         dest::String    #directory
-        target::String  #file inside the archive to test for existence (or blank to check for a.tgz => a/)
+        target::String  #file or directory inside the archive to test
+                        #for existence (or blank to check for a.tgz => a/)
     end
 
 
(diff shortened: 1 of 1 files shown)


tkelman:
cross fingers

IainNZ:
Safe to tag then?

tkelman:
I doubt it could get any worse? julia.mit.edu is down again so I can't easily test on Linux right now without waiting in a travis queue.

tkelman:
btw thanks again @kbarbary for the responsiveness and (knock on wood) fix here

IainNZ:
I'm happy with that logic. Will tag before midnight EST.

kbarbary:
Sure, sorry for the original hasty change!

PR #160, opened 2015-07-29
Fix for Unix which is not Linux

matthb2:
On, for example, FreeBSD, BinDeps calls fail because "EXTENSIONS" is not set. I think this should fix it, at least for operating systems which Julia recognizes as "UNIX" (currently, just Linux and FreeBSD I think).

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -433,9 +433,9 @@ function generate_steps(dep::LibraryDependency, h::Autotools,  provider_opts)
     steps
 end
 
+@unix_only const EXTENSIONS = ["",".so"]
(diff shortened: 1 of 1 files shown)


tkelman:
Odd, this bit of code has been around for some time and apparently no one has noticed? This would give a `WARNING: redefining constant EXTENSIONS` on osx, since that counts under `@unix_only`.

tkelman:
How about just `const EXTENSIONS = ["", Libdl.dlext]` ? That'll come from Compat on 0.3.

matthb2:
@tkelman Libdl.dlext seems to do the right thing on my BSD box and seems like a better solution to me.

PR #161, opened 2015-08-05
fix beginswith deprecation warning

rened:


tkelman:
This is a duplicate of #138 and #149, so I'll ask here what I asked there - have you tested that this is still working fine on packages that trigger these code paths on both 0.3 and 0.4?

rened:
I tested it just now with the following code on 0.3 and current 0.4 on Ubuntu using `./julia --code-coverage=all ~/testbindeps.jl`:
```jl
allpackages = Pkg.available()

packages = Any[]
for x in allpackages
    filename = joinpath(Pkg.dir(x),"REQUIRE")
    if isfile(filename) && ismatch(r"BinDeps", readall(filename))
        push!(packages, x)
    end
end
@show packages

for x in packages
    Pkg.rm(x)
    Pkg.add(x)
end
```
Don't see any problems related to this PR. Coverage info:
```jl
        - function available_versions(p::AptGet)
       48     vers = ASCIIString[]
       48     lookfor_version = false
       48     for l in eachline(`apt-cache showpkg $(p.package)`)
    12732         if startswith(l,"Version:")
        0             try
        0                 vs = l[(1+length("Version: ")):end]
        0                 push!(vers, vs)
        -             end
    12732         elseif lookfor_version && (m = match(DEBIAN_VERSION_REGEX, l)) != nothing
       56             m.captures[2] != nothing ? push!(vers, m.captures[2]) :
        -                                        push!(vers, m.captures[4])
    12676         elseif startswith(l, "Versions:")
       42             lookfor_version = true
    12634         elseif startswith(l, "Reverse Depends:")
       42             lookfor_version = false
        -         end
        -     end
       48     return vers
        - end
```
This did not test `pacman` and `yum` of course - but I think we can be reasonably sure that this works.

tkelman:
`Pkg.rm` followed by `Pkg.add` doesn't usually do anything, it just moves the package to and from `.trash`. Forcing a `git clean` then `Pkg.build` might be a bit more reliable.

PR #171, opened 2015-09-16
String -> AbstractString

yuyichao:
Fix deprecation warnings...

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.3
 URIParser
 SHA
-Compat 0.4.0
+Compat 0.7.1
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -1,7 +1,7 @@
 module BinDeps
     importall Base
     using Compat
-    
+
     export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
             Choice, Choices, CCompile, FileDownloader, FileRule,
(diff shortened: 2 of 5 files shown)


yuyichao:
May require https://github.com/JuliaLang/Compat.jl/pull/139 for 0.3.

tkelman:
I'm guesing coverage is probably pretty bad here. What code paths might hit the `lower(s::Void,collection)` method? If that relies on a not-yet-tagged version of Compat then it'll have to wait for that version to be tagged and reflected in REQUIRE here.

Want to be careful not to break 0.3 of course.

yuyichao:
Bump Compat requirement to 0.7.1 .

Not sure what's the best way to trigger `lower(s::Void, collection)`. I added a `@test BinDeps.lower(nothing, nothing) === nothing` in the test. Someone more familiar with this could probably come up with better test than this....

yuyichao:
@tkelman Can I take your https://github.com/JuliaLang/julia/pull/13107#issuecomment-140935688 as this PR has been tested? =)

tkelman:
Briefly, but only on 0.4. I trust if this breaks anything on 0.3 you'll be available to fix it.

yuyichao:
> I trust if this breaks anything on 0.3 you'll be available to fix it.

Fine. I'll keep an eye on it while I'm figuring out how to get PkgEval working.

PR #173, opened 2015-09-18
Union() -> Union{}

yuyichao:


src/dependencies.jl
```diff
@@ -263,7 +263,7 @@ provider(::Type{Binaries},uri::URI; opts...) = RemoteBinaries(uri)
 provider(::Type{Binaries},path::AbstractString; opts...) = CustomPathBinaries(path)
 provider(::Type{SimpleBuild},steps; opts...) = SimpleBuild(steps)
 provider{T<:BuildProcess}(::Type{BuildProcess},p::T; opts...) = provider(T,p; opts...)
-provider(::Type{BuildProcess},steps::Union(BuildStep,SynchronousStepCollection); opts...) = provider(SimpleBuild,steps; opts...)
+@compat provider(::Type{BuildProcess},steps::Union{BuildStep,SynchronousStepCollection}; opts...) = provider(SimpleBuild,steps; opts...)
```
tkelman: this compat rule was already in the REQUIRE'd version, right? again would be good to be absolutely positive the tests hit this code path on 0.3 too

yuyichao: You mean the version requirement for Compat? The previous PR requires the latest Compat (0.7.1) so I'm pretty sure it's in.

yuyichao: And for this one it should be less problematic (since the rules have been in for longer) and I'm not really sure how can this be reasonably tested. This is why I'd like to tag a new version and fix it if people complains..... ;-p

tkelman: ok

PR #178, opened 2015-09-22
Add filename option for NetworkSource and RemoteBinaries

### LINKED ISSUE #177: Sources provider does not allow to set file name
I'm an author of [MeCab.jl](https://github.com/chezou/MeCab.jl).

In MeCab.jl, we use BinDeps to build from source.
Recently, MeCab changed hosting service from google code to github. But the archived source is available only from Google drive. Google drive's URI is without extension, so we should set file name of archived source file.

I know we can use `download_cmd` with filename option https://github.com/chezou/MeCab.jl/pull/9 , but it seems it is hard coded of file name in Source provider https://github.com/JuliaLang/BinDeps.jl/blob/master/src/dependencies.jl#L323-L325

chezou:
In order to fix #177, I add filename option.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -320,7 +320,7 @@ function generate_steps(dep::LibraryDependency,h::Pacman,opts)
     end
 end
 function generate_steps(dep::LibraryDependency,h::NetworkSource,opts)
-    localfile = joinpath(downloadsdir(dep),basename(h.uri.path))
+    localfile = joinpath(downloadsdir(dep),get(opts,:filename,basename(h.uri.path)))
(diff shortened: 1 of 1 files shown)


tkelman:
I guess this should be okay?

chezou:
Is there any problem or please merge it?

tkelman:
Was waiting to see if anyone else had any feedback, but looks like there hasn't been any.

chezou:
@tkelman thanks for merge it!

chezou:
When released including this commit, I will merge my fix https://github.com/chezou/MeCab.jl/pull/9

PR #179, opened 2015-09-24
Don't touch deps.jl at all if do_install is false

staticfloat:
@Keno could you please take a look at this to see if it's in the spirit of the `do_install` flag?  Essentially, without this, when creating a `PackageContext` with `do_install` set to `false` (as [happens in `BinDeps.debug`](https://github.com/JuliaLang/BinDeps.jl/blob/1803cfc2c0b2a044b2753c8e634c2775185f0329/src/debug.jl#L51-L54)) `deps.jl` gets clobbered, wiping out the build configuration.  With this change, `deps.jl` doesn't get touched at all when `do_install` is set to `false`, which (as far as I can tell) is what is the intended behavior.  If there's some other circumstance I'm not thinking of, please let me know. [Here's one case](https://github.com/JuliaLang/BinDeps.jl/blob/1803cfc2c0b2a044b2753c8e634c2775185f0329/src/dependencies.jl#L889) where `do_install` is set to `false` (and I believe this case would also benefit from this change), but in [the last case](https://github.com/JuliaLang/BinDeps.jl/blob/1803cfc2c0b2a044b2753c8e634c2775185f0329/src/dependencies.jl#L959) where `do_install` is set to `false`, I'm less confident.  Thanks.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -795,33 +795,33 @@ macro install(_libmaps...)
                                 end
                             end
                         end
-                    end
 
-                    # Generate "deps.jl" file for runtime loading
-                    depsfile = open(joinpath(splitdir(Base.source_path())[1],"deps.jl"), "w")
-                    println(depsfile,
-                        """
-                        # This is an auto-generated file; do not edit
-                        """)
-                    println(depsfile, "# Pre-hooks")
-                    println(depsfile, join(pre_hooks, "\n"))
-                    println(depsfile,
-                        """
-                        # Macro to load a library
-                        macro checked_lib(libname, path)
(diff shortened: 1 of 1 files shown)


Keno:
You're right about the first case. The second case is a little ambiguous. It exists to force building with a particular method. Arguably it should write a `deps.jl` file, but only considering the passed in method. I think for now it's fine not to generate the deps.jl file.

staticfloat:
Wouldn't we want `do_install` set to `true` in that case?  Unless you want it to write `deps.jl` without actually installing dependencies?  So in that case we'd need a `do_install` flag as well as a `write_depsjl` flag.

Keno:
The problem is that as written the logic in the install macro looks at
every method not just the given one. Originally the install macro would
just call satisfy! which is what the build function then does manually (but
only with the selected method). Over time more stuff got added to the
install macro that probably should have been reflected there. In any case,
the reps.jl file it currently creates is probably broken, so this can't be
worse. Fixing it properly would involve refactoring the install macro.

staticfloat:
So...... merge when green?

PR #185, opened 2015-10-31
Fix dlopen c api change

### LINKED ISSUE #184: julia0.5 dlopen api change
PR https://github.com/JuliaLang/julia/pull/13796 caused issues for BinDeps where it does a direct ccall to a function that no longer exists in base at https://github.com/JuliaLang/BinDeps.jl/blob/e35913637c1c30149bcc7bdba5bd5ebede711efa/src/dependencies.jl#L510-L519l

yuyichao:
Caused by JuliaLang/julia#13796

Also fix tests when the build is running in a separate process and BinDeps is not pulled in at runtime.

No additional tests added since the current test already trigger this path on my computer and it seems a little hard to trigger this path directly otherwise.

This shouldn't break 0.3 or 0.4 since everything new should be behind a version check.

Fix #184

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -505,16 +505,28 @@ function _find_library(dep::LibraryDependency; provider = Any)
     return ret
 end
 
-function check_path!(ret,dep,opath)
-    flags = Libdl.RTLD_LAZY
-    handle = Libc.malloc(2*sizeof(Ptr{Void}))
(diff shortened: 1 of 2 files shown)


tkelman:
lgtm, thanks.

tkelman:
I'll tag this since I hope it won't cause any issues on 0.3 or 0.4.

PR #186, opened 2015-11-02
Fix dlopen c api change #184

Ph0non:
Fix referred in https://github.com/JuliaLang/BinDeps.jl/issues/184 still throws an error. Suggestes solution from @Carreau works fine.


tkelman:
Thanks, it's green so let's try it.

PR #187, opened 2015-11-16
Prepare BinDeps for upcoming SHA.jl changes

staticfloat:
`SHA` is going to return bytes instead of strings in the future

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -983,6 +983,11 @@ using SHA
 function sha_check(path, sha)
     open(path) do f
         calc_sha = sha256(f)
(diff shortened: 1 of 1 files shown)


src/dependencies.jl
```diff
@@ -983,6 +983,9 @@ using SHA
 function sha_check(path, sha)
     open(path) do f
         calc_sha = sha256(f)
+        if !isa(calc_sha, @compat AbstractString)
```
nalimilan: Actually, you don't even need to write `@compat` for types, which are provided as aliases.

Also, a small comment to know when this can happen would be useful. It will allow removing this hack easily in the future.

staticfloat:
Done and done.

PR #189, opened 2015-12-02
Test add zypper for openSUSE to bindeps dependencies

colbec:
Added some code for zypper as a package manager


tkelman:
I think this is redundant with #193, so I'll close.

Issue #188, opened 2015-12-01
Any plans to add openSUSE zypper/Yast to the list of package managers?


colbec:
Using openSUSE Leap 42.1:
Just wondering if there is any argument why zypper might not be on the list of things to add.
Right now, if Julia finds it has to recompile stale cache files for PyPlot for example, there is a strange call which opens up a window for Yast package manager wanting root access, presumably to pull in or modify existing binaries. Dismissing this window allows the process to continue, and PyPlot appears to work correctly at least on simple plots.

tkelman:
Nobody has added it yet, but I'd be happy to review and help anyone who'd like to try writing a pull request to add support for opensuse packages here. It's quite a nice distro, really great build service and we use its cross compiled packages on windows, but I think it's a little less popular than debbuntu, fedora, and arch. Natural to add support for it though.

I think PyPlot should be using Conda.jl for its binary deps these days, it might be something else like Nettle.jl or ZMQ.jl causing the yast popup if you're using IJulia. Try Pkg.build for a few packages to see if you can identify which one it is.

colbec:
Agreed, code only needs to be added as and when there is a demand, and for less popular distros the additions come a bit later. I have posted a request at the openSUSE forums to see if there is any previous work, but no response so far. The science repo has a working version of v4 so there is a chance that someone has already beaten this bush already.

I tested 'build' on the suggested packages but none of them triggered the yast issue. I will keep trying the growing list of installed packages. IJulia no longer seems to be causing the yast popup, although it was a few days ago.

I'm not really familiar with git pulls or Julia programming but I am willing to give it a go. I'm assuming that I just have to make mods to file dependencies.jl using say pacman as a template. We'll see how it goes.

colbec:
I have opened a pull request at https://github.com/colbec/BinDeps.jl
Not sure if it has landed in the right place, so I am offering the link here.

colbec:
OK, puzzled by the way my detection of zypper is progressing.
I have inserted some test code which shows that my zypper is found and the flags are set.
The puzzle is that the Yast window pops up before the code apparently starts looking for it.
So I'm stuck on two things:
1. something in the code is triggering Yast before it is asked for
2. when I try to display the name of the package that the process is looking for ($(?.package)) the compiler tells me that ? is not defined.

nalimilan:
What is `$(?.package)` supposed to do? That's a very weird syntax, which would work only if the variable `?` holds an object with a `package` field.

[...]

PR #190, opened 2015-12-03
RP 189 adding zypper to BinDeps updated

colbec:


src/dependencies.jl
```diff
@@ -199,6 +199,30 @@ pkg_name(p::Pacman) = p.package
...
+type Zypper <: PackageManager
+    package::AbstractString
+end
+can_use(::Type{Zypper}) = has_zypper && OS_NAME == :Linux
+package_available(z::Zypper) = can_use(Zypper) && success(`zypper se $(z.package)`)
+function available_version(z::Zypper)
+    uname = readchomp(`uname -m`)
+    found_uname = false
+    found_version = false
+    for l in eachline(`zypper info $(z.package)`)
```
tkelman: ref https://github.com/JuliaLang/BinDeps.jl/issues/188#issuecomment-161597464, I think we want to do something like
```
ENV2 = copy(ENV)
ENV2["LC_ALL"] = "C"
for l in eachline(setenv(`zypper info $(z.package)`, ENV2))
```

And what are `found_uname` and `found_version` being used for here?

colbec:
Oh dear, the variables `found_*` are unintentional residual from testing that should have been removed.
Issue https://github.com/JuliaLang/julia/issues/14249 which has a bearing on the `ENV` variable setting seems to be still ongoing. Proceed anyway?

tkelman:
Whatever the resolution of that issue will be, it will take effect in 0.5 at the earliest. BinDeps.jl still needs to support 0.3 and 0.4 at the moment. I would hope any new features will have deprecation paths, at which point Compat.jl and BinDeps.jl could be updated accordingly, but for now I think you should proceed if you'd like to see this feature in the short term.

nalimilan:
Agreed, that discussion shouldn't block you from adding Zypper support.

tkelman:
I think this is redundant with #193, so I'll close.

PR #191, opened 2015-12-04
Fix detection of PowerShell on windows

davidanthoff:
``powershell --help`` throws an error, the correct way to make that call is ``powershell -h``.

Without this fix there is always a red error output when e.g. Ipopt ``build`` is run.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -46,10 +46,15 @@ module BinDeps
     function download_cmd(url::AbstractString, filename::AbstractString)
         global downloadcmd
         if downloadcmd === nothing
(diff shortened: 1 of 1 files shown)


tkelman:
I think all 4 of these commands could alternately accept just `-h`, but this should work fine.

davidanthoff:
Would also make sense to suppress the output, really no reason to show the Powershell help during this :) I tried with piping to the null device, but didn't succeed...

tkelman:
The output should be suppressed? None of the BinDeps-using packages I've run locally or on appveyor show any output from this check.

davidanthoff:
I opened a new issue #192.

Issue #192, opened 2015-12-05
On windows powershell help is displayed


davidanthoff:
On master I get this

````
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.4.1 (2015-11-08 10:33 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-w64-mingw32

julia> Pkg.build("Ipopt")
INFO: Building WinRPM
INFO: Downloading https://cache.e.ip.saba.us/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_13.2/repodata/repomd.xml
INFO: Downloading https://cache.e.ip.saba.us/http://download.opensuse.org/repositories/windows:/mingw:/win64/openSUSE_13.2/repodata/repomd.xml
INFO: Building Ipopt

PowerShell[.exe] [-PSConsoleFile <file> | -Version <version>]
    [-NoLogo] [-NoExit] [-Sta] [-Mta] [-NoProfile] [-NonInteractive]
    [-InputFormat {Text | XML}] [-OutputFormat {Text | XML}]
    [-WindowStyle <style>] [-EncodedCommand <Base64EncodedCommand>]
    [-File <filePath> <args>] [-ExecutionPolicy <ExecutionPolicy>]
    [-Command { - | <script-block> [-args <arg-array>]
                  | <string> [<CommandParameters>] } ]

PowerShell[.exe] -Help | -? | /?

-PSConsoleFile
    Loads the specified Windows PowerShell console file. To create a console
    file, use Export-Console in Windows PowerShell.

-Version
    Starts the specified version of Windows PowerShell.
    Enter a version number with the parameter, such as "-version 2.0".

-NoLogo
    Hides the copyright banner at startup.

-NoExit
    Does not exit after running startup commands.

-Sta
    Starts the shell using a single-threaded apartment.
    Single-threaded apartment (STA) is the default.

-Mta
    Start the shell using a multithreaded apartment.

-NoProfile
    Does not load the Windows PowerShell profile.

-NonInteractive
    Does not present an interactive prompt to the user.

-InputFormat
    Describes the format of data sent to Windows PowerShell. Valid values are
[...]

PR #194, opened 2015-12-20
Prefix powershell with cmd to fix reduced shell command set in J 0.4.2.

hustf:
This secondarily "fixes" https://github.com/JuliaLang/BinDeps.jl/issues/192 by reducing unnecessary text output from Windows powershell. 

* The fixed issue may not appear on all Windows systems. This I assume depends on the context of ENV["PATH"], which is shared by many Windows 10 programs and shouldn't need to have special settings for Julia.

* Pkg.test("BinDeps") still fails due to an error in "HttpParser" ('deps.jl' is missing, but needed).

* The 'cmd' prefix has not been combined with the other download engines, as I'm unable to test it.

*  This ought to be tested on various Windows systems.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -48,9 +48,9 @@ module BinDeps
         if downloadcmd === nothing
             for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
-                    checkcmd = `$download_engine -help`
-                else
-                    checkcmd = `$download_engine --help`
+                    checkcmd = `cmd $download_engine new-variable foo`
+                 else
+                    checkcmd = `cmd $download_engine --help`
                 end
                 try
                     if success(checkcmd)
@@ -69,7 +69,7 @@ module BinDeps
         elseif downloadcmd == :fetch
             return `fetch -f $filename $url`
         elseif downloadcmd == :powershell
-            return `powershell -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
+            return `cmd powershell -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
         else
             error("No download agent available; install curl, wget, or fetch.")
         end


src/BinDeps.jl
```diff
@@ -48,9 +48,9 @@ module BinDeps
         if downloadcmd === nothing
             for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
-                    checkcmd = `$download_engine -help`
-                else
-                    checkcmd = `$download_engine --help`
+                    checkcmd = `cmd $download_engine new-variable foo`
+                 else
+                    checkcmd = `cmd $download_engine --help`
```
tkelman: this is wrong on unix. also the change in indent on the preceding line is superfluous

src/BinDeps.jl
```diff
@@ -48,9 +48,9 @@ module BinDeps
         if downloadcmd === nothing
             for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
-                    checkcmd = `$download_engine -help`
-                else
-                    checkcmd = `$download_engine --help`
```
tkelman: why the change from `-help` to `new-variable foo` ?

davidanthoff:
What is the primary problem this is fixing?

hustf:
>this is wrong on unix
Of course. The obvious workaround would be to replace :powershell with something like parse("cmd powershell"), which doesn't work. I'd like to pursue your other questions first before posting a new proposal or removing the "feature" of using cmd. And I have these related questions: Does Julia currently support any Windows systems that do not have  Powershell in  %SystemRoot%\system32\WindowsPowerShell\v1.0\powershell.exe? 

If not, do curl, wget and fetch add flexibility at all?

>also the change in indent on the preceding line is superfluous
Noted, will be fixed if we find a good solution.

hustf:
> why the change from -help to new-variable foo ?

This secondarily "fixes" #192 ->  #14314 by reducing unsuppressable text output from Windows powershell under  WMF 5 (whatever that is). Nobody likes screen clutter. Installation time will also be slightly reduced.

davidanthoff:
I still don't understand why ``powershell`` needs to be prefixed with ``cmd`` in the first place?

davidanthoff:
> This secondarily "fixes" #192 -> #14314 by reducing unsuppressable text output from Windows powershell under WMF 5 (whatever that is). Nobody likes screen clutter. Installation time will also be slightly reduced.

Yes, I like that particular change!

davidanthoff:
Btw., we should probably also add the ``-NoProfile`` option whenever we call powershell, really no need to run user scripts when all we want to do is download something.

hustf:
> What is the primary problem this is fixing?

[ Sorry, I meant to link here in the original submittal. 
https://groups.google.com/forum/#!topic/julia-users/pHonnw8X3iw
A post Dec. 20 introduces the 'cmd powershell' workaround, and the wider problem is described at great length earlier in the thread. ]

The primary problem this is fixing: BinDeps is not working at all on my system without it. I have met this wall when installing other packages and libraries, among them Cairo.

The PATH variable on my computer seems to be modified by various auto-updating software. Cleaning PATH requires a restart every time, and also administrator privileges as far as I can see. It is not even visible in the control panel for users without administrator rights. So the number of experiments I have tried with PATH modifications is low. Better to not thrust it in my opinion, and rely only on executables directly in System/ System32.

davidanthoff:
Hm, this makes little sense to me. The PowerShell directory (``%SYSTEMROOT%\System32\WindowsPowerShell\v1.0\``) should be on the PATH, the Windows setup puts it there, lots of things will break on a system where that isn't the case. I don't think we should add workarounds for systems that e.g. remove that directory from PATH. Are you trying to remove the PowerShell directory manually from PATH? Or is it on your PATH, but things are still not working?

But even more confusing: if the PowerShell directory is NOT on PATH, why would PowerShell start if called via the ``cmd`` indirection?? ``cmd`` also just looks for it in the directories that are on PATH. So if the problem is that PowerShell is not on the path, it should also not work via ``cmd``...

To me this all sounds more like your system is somehow in a weird state, rather than a bug in BinDeps here...

hustf:
Davidantoff, thanks for the input; I agree this seems weird. You should know I just saw even the first of your replies now, and I also edited a bit after posting.  Just to make sure we're not talking past each other here - you might look briefly over again. I will certainly do a bit more twiddling with my PATH settings, just not in the first few days. My Windows system is a partition of a Mac using 'Bootcamp', and I have had three versions of Visual Studio on it.

davidanthoff:
I've created a PR #195 that has the change that I think should go in, plus the ``-NoProfile`` thing I mentioned. Both of those seem like no-brainers to me, unlike this PR here, which needs more research.

hustf:
That`s nice, I'll close it after some days for possible further comments.

I just installed Julia on a new laptop with Windows 10 and nothing else. Shell commands like echo, dir etc. don't work but cmd and powershell work (included in the path, and early in it). That means bindeps works.

For possible reference, I include default PATH on a clean Windows 10. 

````
Cmd->Echo %Path%
C:\Program Files (x86)\Intel\iCLS Client\;C:\Program Files\Intel\iCLS Client\;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\DAL;C:\Program Files\Intel\Intel(R) Management Engine Components\DAL;C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\IPT;C:\Program Files\Intel\Intel(R) Management Engine Components\IPT;C:\Program Files\Intel\WiFi\bin\;C:\Program Files\Common Files\Intel\WirelessCommon\
````



````
julia> ENV["PATH"]
"C:\\Users\\F\\AppData\\Local\\Julia-0.4.2\\bin;C:\\Users\\F\\AppData\\Local\\Julia-0.4.2\\bin\\..\\Git\\bin;C:\\Program Files (x86)\\Intel\\iCLS Client\\;C:\\Program Files\\Intel\\iCLS Client\\;C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\;C:\\Program Files (x86)\\NVIDIA Corporation\\PhysX\\Common;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\DAL;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\DAL;C:\\Program Files (x86)\\Intel\\Intel(R) Management Engine Components\\IPT;C:\\Program Files\\Intel\\Intel(R) Management Engine Components\\IPT;C:\\Program Files\\Intel\\WiFi\\bin\\;C:\\Program Files\\Common Files\\Intel\\WirelessCommon\\"
````


http://sourceforge.net/p/msys2/wiki/MSYS2%20introduction/
>
When using the shells, try to remove as many entries from PATH as you can, ideally only leaving something like C:\Windows\system32. Mixing in programs from other MSYS2 installations, Cygwin installations or compiler toolchains is not supported and will probably break things in unexpected ways. Do not have these things in PATH when running MSYS2 unless you know exactly what you're doing.

PR #195, opened 2015-12-24
Improve PowerShell calls in download_cmd

davidanthoff:
Two changes: 1) Change the call to PowerShell that is meant to detect whether PowerShell is on the system to not emit any output, 2) always specify the -NoProfile argument so that user PowerShell
startup scripts are not run.

Change 1) was suggested in #194.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -49,7 +49,7 @@ module BinDeps
         if downloadcmd === nothing
             for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
-                    checkcmd = `$download_engine -help`
+                    checkcmd = `$download_engine -NoProfile -Command ""`
                 else
                     checkcmd = `$download_engine --help`
                 end
@@ -70,7 +70,7 @@ module BinDeps
         elseif downloadcmd == :fetch
             return `fetch -f $filename $url`
         elseif downloadcmd == :powershell
-            return `powershell -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
+            return `powershell -NoProfile -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
         else
             error("No download agent available; install curl, wget, or fetch.")
         end


tkelman:
running the julia `success` function should not be producing any output. there's a bug, but it's not here. needs to be looked at in more detail to find the actual fix, not shoved under the rug like this.

Varanas:
Why is that a bug, @tkelman? `success` isn't specified to throw away output of the spawned process -- it's  implemented with [`wait(spawn(cmd))`](https://github.com/JuliaLang/julia/blob/79b241872d55d556189ce2fb4e91157164b2b1a6/base/process.jl#L610-L617) so I wouldn't really *expect* it to do that.

tkelman:
Without any additional inputs, `spawn` doesn't do anything with the output, just runs the process.

Varanas:
So this isn't supposed to happen (aside from the screwed up printing, I mean)?
spawn

davidanthoff:
@tkelman This PR is not preventing anyone from fixing the underlying bug in ``success``, but it does fix a pretty confusing UI weirdness for users on Windows that is going to hit anyone with PowerShell 5 (i.e. all Win10 users, and anyone else who updates PowerShell on down level systems). Unless we know someone who can and will (i.e. has the time and capability) fix the underlying bug fairly soon, I would suggest to just merge this here. Especially since this PR here has no downside, i.e. the code is in no way "less good" than the previous version. There really seems no cost involved in this workaround.

Also, the ``-NoProfile`` option should be merged no matter what, not having that is simply a bug.

tkelman:
I don't think "new-variable foo" is the most obvious test for whether powershell is present.

davidanthoff:
@tkelman Better?

davidanthoff:
@tkelman Bump, any chance this could get merged?

tkelman:
We can add the `-NoProfile`, but would rather find and fix the bug that's causing things to output that shouldn't be, rather than change the detection code.

davidanthoff:
I suspect that it might be very difficult to fix the output issue. I looked into this, and my best guess is that Powershell 5 switched from using the high-level console input and output functions to the low-level functions. The current redirection support in libuv for console redirection only works for programs that use the high-level console functions, as far as I can tell. I think any console program that uses the low-level console API will show the same behavior of showing output, even if it is redirected, because that is simply not supported in libuv. My pretty hasty research on how difficult it is to fix this suggested that it might be almost [impossible](http://homepage.ntlworld.com/jonathan.deboynepollard/FGA/capture-console-win32.html) to get this right.

But at the end of the day I really don't understand the reasoning to not just merge both: this fixes a UI bug (clearly showing all PowerShell help when we updated any package that uses BinDeps is a UI bug, right?). This bug fix here has been sitting here for two months now, clearly no one has the time right now to fix the underlying issue, so why not merge this and make julia a little better for users in the meantime?

tkelman:
Because it's completely harmless? It also works as intended when Julia is started within mintty, so I don't think it's impossible to fix.

src/BinDeps.jl
```diff
@@ -48,7 +48,7 @@ module BinDeps
         if downloadcmd === nothing
             for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
-                    checkcmd = `$download_engine -help`
+                    checkcmd = `$download_engine -NoProfile -Command "\$no_op"`
```
tkelman: The `"\$no_op"` is ugly and confusing. Just change this to `-Command ""` and I'll merge it.

davidanthoff:
@tkelman Thanks for the suggestion, that makes the thing much cleaner.

davidanthoff:
Done.

PR #197, opened 2016-01-26
Add tar.Z file support.

helgee:
`tar.Z` files can be unpacked with the same flags used for `tar.gz` files.


tkelman:
Have you tested this locally? Do you have a package that will use this regularly enough to notice if it breaks?

helgee:
Yes, I tested it on OSX and Linux. I am developing a wrapper for the [NAIF SPICE toolkit](http://naif.jpl.nasa.gov/naif/). The sources are distributed in tar.Z format.

helgee:
It also works on Windows.

PR #198, opened 2016-01-27
Turn on precompilation

yuyichao:
This should fix the intermittent package loading failure on many packages that depends on BinDeps (Notably ZMQ and IJulia).

If it is unsafe to turn precompilation on for this package, we should instead explicitly specify it to avoid the intermittent failure.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -1,3 +1,4 @@
+isdefined(Base, :__precompile__) && __precompile__()
 module BinDeps
     importall Base
     using Compat


src/BinDeps.jl
```diff
@@ -1,3 +1,4 @@
+VERSION >= v"0.4.0" && __precompile__()
```
tkelman: `isdefined(Base, :__precompile__)` to be more specific and also address the problem for later prereleases?

yuyichao: done

PR #200, opened 2016-01-31
RFC: make sure aliases are always used as an array of strings

tkelman:
to avoid the ```MethodError: `joinpath` has no method matching joinpath(::UTF8String, ::Char)```
seen at https://github.com/jiahao/GSL.jl/pull/39

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -413,6 +413,10 @@ function gethelper(dep::LibraryDependency,method)
     return (nothing,nothing)
 end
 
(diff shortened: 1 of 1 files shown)


nalimilan:
LGTM, though you could also have restricted the second signature to `AbstractArray`.

tkelman:
Thanks. Maybe someone could want to input an iterator of alias names or something?

nalimilan:
Yeah, there's little risk in keeping the broad signature. In the worst case, an error would likely be raised later anyways.

PR #201, opened 2016-02-08
Fix method overwritten warning on 0.5-dev

yuyichao:
This method is defined in `src/debug.jl`


tkelman:
Might be a good, relatively calm opportunity to tag a new version. Has been a little while.

PR #203, opened 2016-02-11
Enable AppVeyor

tkelman:
Can I get a volunteer to run this on their account? My account is crowded and already has half-a-dozen daily builds for various packages, would like to spread out the queue time if anyone's willing.


nalimilan:
I've just set it up on my account:
https://ci.appveyor.com/project/nalimilan/bindeps-jl

I never know whether the Webhook is needed or not, so here it is: https://ci.appveyor.com/api/github/webhook?id=v4h9hftbg11fst1n

tkelman:
Thanks! Let's see. We can add the fast fail powershell snippet if it becomes necessary.

PR #204, opened 2016-02-11
Revert "Allow providing multiple package names as vector in provides()"

tkelman:
This broke WinRPM, ref https://github.com/JuliaLang/Cairo.jl/issues/132

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ defined the following dependencies:
 ```
 
 Let's suppose that these libraries are available in the `libfoo-dev` and `libbaz-dev`
-in apt-get and that both libraries are installed by the `baz` or the `baz1` yum package, and the `baz` pacman package. We may
+in apt-get and that both libraries are installed by the `libbaz` yum package and the `baz` pacman package. We may
 declare this as follows:
 
 ```julia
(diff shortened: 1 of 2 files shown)


tkelman:
One of the Travis nightly jobs hit a `The command "curl -s -L -o julia.dmg 'https://s3.amazonaws.com/julianightlies/bin/osx/x64/julia-latest-osx.dmg'" failed and exited with 18 during .`, restarted.

nalimilan:
I don't understand why this fails, but a possible workaround would be to add this to WinRPM:
```julia
function provides(::Type{RPM}, packages::AbstractArray, dep::LibraryDependency; opts...) = provides(provider(RPM, p; opts...), dep; opts...)
```

This would effectively revert this change only for WinRPM, which is the only provider to support passing multiple packages on its own (i.e. even without my PR). Then we could look for a long-term fix. Could you check whether it works?

tkelman:
Will try that.

tkelman:
Yep, that works, good call.

nalimilan:
Cool, but we should investigate why this doesn't work.

tkelman:
Agreed.

PR #205, opened 2016-03-06
Use powershell -v 2 to avoid excess output

### LINKED ISSUE #192: On windows powershell help is displayed

tkelman:
fixes #192, closes #195

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -49,7 +49,7 @@ module BinDeps
         if downloadcmd === nothing
             for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
-                    checkcmd = `$download_engine -help`
+                    checkcmd = `$download_engine -NoProfile -v 2 -help`
                 else
                     checkcmd = `$download_engine --help`
                 end
@@ -70,7 +70,7 @@ module BinDeps
         elseif downloadcmd == :fetch
             return `fetch -f $filename $url`
         elseif downloadcmd == :powershell
-            return `powershell -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
+            return `powershell -NoProfile -v 2 -Command "(new-object net.webclient).DownloadFile(\"$url\", \"$filename\")"`
         else
             error("No download agent available; install curl, wget, or fetch.")
         end


davidanthoff:
So this is probably a little less robust than what we had previously: one can uninstall powershell V2 mode from Windows, it is an optional component (but installed by default on consumer editions, not sure about server editions). The old approach of targeting powershell without a version identifier should always work because one can't remove all powershell versions from modern versions of Windows, as far as I know.

nalimilan:
Maybe try with V2 and if if fails fall back to the old command?

davidanthoff:
Or just use #195... The argument against that PR was that it didn't fix the underlying problem with the ``success`` function. Clearly this PR here also doesn't. Both are hacks around the underlying problem. But #195 seems more robust because it will work on more versions of Windows.

PR #208, opened 2016-03-28
Add GSL.jl example and update link to Cairo.jl in README.md

nalimilan:
A simple example is easier to understand for starters.
The link to Cairo.jl pointed to an old branch, better follow master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,11 @@ the actual package, I want to answer a few common questions:
 
 # The high level interface - Declaring dependencies
 
-To get a feel for the high level interface provided by BinDeps, have a look at a 
-real-world example, namely the [build script from the Cairo package](https://github.com/JuliaLang/Cairo.jl/blob/kf/bindeps2/deps/build.jl).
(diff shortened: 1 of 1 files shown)


nalimilan:
GSL.jl is the only package I could find with the simple case, and yet support for all platforms, as well as both source and binaries. Other suggestions welcome.

tkelman:
We may want to point to specific tags as opposed to master, since these may change.

nalimilan:
Yes, I wondered about that, but in the end I concluded that we'd better always refer to master. As this PR shows, the links will likely get out of date if we choose specific tags, which would really be awful for documentation purposes. OTC, master should always reflect the current status of BinDeps.

tkelman:
Not if some replacement for BinDeps comes along and those packages start using it?

nalimilan:
Well, yeah, but if this ever happens it will be only once and we can fix it. OTC, smaller changes/improvements to the BinDeps syntax are more likely to be frequent. Giving examples that don't work would be terrible.

nalimilan:
So, final word? :-)

PR #209, opened 2016-03-29
unpack_cmd for .tar files under windows

greenflash1357:
If a .tar file is not further compressed, it wasn't possible to unpack it under windows.


tkelman:
Seems reasonable. Have you tested this locally?

greenflash1357:
yes. It works for me.

PR #211, opened 2016-04-20
Check for `apt-cache` as well, as it's necessary for apt provider

staticfloat:
This is useful in case you have a system like OpenSUSe where you can install `apt-get` for some reason, but `apt-cache` doesn't come with it; you still want to fallback to `Zypper`.


tkelman:
That sounds like there's a fun story behind it...

Issue #213, opened 2016-04-27
Dependency on executables


iamed2:
Sometimes a package depends on an executable file, not a library. 

While package providers can install dependencies in the same manner, `find_library` would need to be substituted with an appropriate `find_executable` function (e.g. "an executable is installed if `which` or `where` <exec_name> returns a file or it's in deps/usr/bin" or something like that). `deps.jl` would maybe also set the path of the executable in a variable. There would then be `executable_dependency` in place of `library_dependency`. 

I'm keen to write this and submit a PR unless anyone has any suggestions or objections.

nalimilan:
Makes sense to me. Like for libraries, you may also want to accept a `validate` function which would be called with the path to the executable, and return a boolean indicating whether the executable fits the bill.

iamed2:
1) I didn't know that existed, cool
2) Definitely!

tkelman:
Sounds like a good idea, would be happy to review something like this. One thing to note is that `which` isn't always installed by default on e.g. minimal Fedora containers, so ideally there should be some sort of fallback - "`find_executable` cannot work without having `which` installed" or something. Manually crawling `PATH` sounds like more trouble than it's worth.

iamed2:
Would `find` be installed? Could use that to crawl `PATH`.

nalimilan:
If `which` isn't available, do we at least have a way of trying to run the command using only its name, e.g. via `/bin/sh`? Else, we could simply act as if the executable was missing, i.e. build the sources, so that we know where the binary is.

tkelman:
`/usr/bin/env foo` maybe? depends whether executing with no arguments is expected to work meaningfully

iamed2:
We could have the `validate` function run even if `which` wasn't installed, and let the user determine what is a valid "no-op" call.

nalimilan:
> We could have the validate function run even if which wasn't installed, and let the user determine what is a valid "no-op" call.

The problem is, `validate` should be able to run the executable very easily, without a special cases. Else, package authors won't test that path (since they don't use weird systems) and it will be broken too often. So `validate` should always be passed one argument which can be handled in a single code path, e.g. by calling `run` on it after adding some additional arguments.

iamed2:
I booted a VM with a minimal Fedora install (no extras, not even the Standard ones) and it came with `which` in `/usr/bin`.

tkelman:
It's docker containers where I've seen it missing and need to be manually installed (to be able to build Julia).

PR #214, opened 2016-04-29
WIP: Depend on executables, implements #213

iamed2:
Users can now declare `executable_dependency` and have it work just like `library_dependency`. 

Unfortunately as far as I can tell there is no easy way in Julia to check if a file is executable on Windows. However, I ensured that files with `PATHEXT` executable extensions are preferred. 

Currently `install` will write a variable `_exec` to deps.jl which contains a dictionary of dependency names and paths to executables. Please let me know if there is a better way. I mostly left `load_dependencies` alone, so it still gives tuples where the second element is the path. It's not clear whether including `deps.jl` or calling `load_dependencies` is preferred, and the results for doing either are undocumented. If someone knows the recommended way to approach this, I can document it. 

It's unclear to me how to test this new feature. Would the best approach be to just create and run a build script with an arbitrary executable dependency? The current approach seems to use build scripts from other packages, but since it's a new feature, no packages are using it. I have been testing with my build script for the unreleased EphemeralPG.

Work left:
- [x] Document `executable_dependency` ~~and the recommended way of loading dependencies~~
- [x] Adapt the Autotools build process for executable dependencies
- [x] Test
- [x] ~~Fix or~~ get around https://github.com/JuliaLang/Compat.jl/issues/187 (currently causing 0.4 failure)

This is here in WIP state so I can get comments sooner rather than later.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -40,8 +40,6 @@ type DependencyGroup
     deps::Vector{Dependency}
 end
 
-typealias LibraryGroup DependencyGroup
-
 # Default directory organization
 pkgdir(dep) = dep.context.dir
 depsdir(dep) = joinpath(pkgdir(dep),"deps")
@@ -55,17 +53,34 @@ srcdir(dep) = joinpath(depsdir(dep),"src")
 libdir(provider, dep) = [libdir(dep), libdir(dep)*"32", libdir(dep)*"64"]
 bindir(provider, dep) = bindir(dep)
 
+destdir(dep::LibraryDependency) = libdir(dep)
+destdir(provider, dep::LibraryDependency) = libdir(provider, dep)
+destdir(dep::ExecutableDependency) = bindir(dep)
+destdir(provider, dep::ExecutableDependency) = bindir(provider, dep)
+
 successful_validate(l, p=nothing) = true
+
+# normal contents of %PATHEXT% on Windows
+const win_pathext = (".COM", ".EXE", ".BAT", ".CMD", ".VBS", ".VBE", ".JS", ".JSE", ".WSF", ".WSH", ".MSC")
+const win_case_pathext = [collect(win_pathext); map!(lowercase, collect(win_pathext))]
 
 function BinDeps.isexecutable(filepath)
     if isfile(filepath)
         res = false
         @unix_only begin
-            res = ccall(:access, Cint, (Ptr{Uint8}, Cint), filepath, 1) == 0
+            res = ccall(:access, Cint, (Ptr{UInt8}, Cint), filepath, 1) == 0
         end
         # unfortunately there is no easy way to tell if this process can execute a file on Windows
         @windows_only begin
-            res = true
+            res = any(map(x->endswith(filepath, x), win_case_pathext))
+            if res && (endswith(filepath, ".exe") || endswith(filepath, ".EXE"))
+                try
+                    # check for the DOS/Windows executable magic number 'MZ'
+                    res = read(filepath, UInt16) == 0x5a4d
+                catch
+                    res = false
+                end
+            end
         end
         if res
             return true
@@ -79,7 +94,7 @@ function BinDeps.isexecutable(filepath)
 end
 
 # constructor for ExecutableDependency and LibraryDependency
-@compat function (::Type{T}){T<:Dependency}(context::PackageContext, name; properties...)
+function _dependency{T<:Dependency}(::Type{T}, context::PackageContext, name; properties...)
     validate = successful_validate
     group = nothing
     for i in 1:length(properties)
@@ -126,12 +141,14 @@ macro setup()
         end
         @deprecate library_group(args...) dependency_group(args...)
         dependency_group(args...) = BinDeps.DependencyGroup(bindeps_context,args...)
-        library_dependency(args...; properties...) = BinDeps.LibraryDependency(
+        library_dependency(args...; properties...) = BinDeps._dependency(
+            BinDeps.LibraryDependency,
             bindeps_context,
             args...;
             properties...
         )
-        executable_dependency(args...; properties...) = BinDeps.ExecutableDependency(
+        executable_dependency(args...; properties...) = BinDeps._dependency(
+            BinDeps.ExecutableDependency,
             bindeps_context,
             args...;
             properties...
@@ -147,9 +164,6 @@ executable_dependency(args...; properties...) = error("No context provided. Did 
 # binary finders
 const has_which = try success(`which which`) catch e false end
 const has_where = try success(`where where`) catch e false end
-
-# normal contents of %PATHEXT% on Windows
-const win_pathext = (".COM", ".EXE", ".BAT", ".CMD", ".VBS", ".VBE", ".JS", ".JSE", ".WSF", ".WSH", ".MSC")
 
 abstract PackageManager <: DependencyProvider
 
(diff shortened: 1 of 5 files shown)


tkelman:
The `load_dependencies` method is much less widely used than including `deps.jl`. The latter has the benefit that BinDeps is not required by the package at runtime, though precompilation mitigates the cost of that somewhat.

src/dependencies.jl
```diff
@@ -15,20 +17,31 @@ type PackageContext
...
 end
 
-type LibraryGroup
+type DependencyGroup
     name::AbstractString
-    deps::Vector{LibraryDependency}
+    deps::Vector{Dependency}
 end
 
+typealias LibraryGroup DependencyGroup
```
nalimilan: I guess this is temporary (because of WIP) ?

src/dependencies.jl
```diff
@@ -42,9 +55,31 @@ srcdir(dep) = joinpath(depsdir(dep),"src")
...
-successful_validate(l,p) = true
+successful_validate(l, p=nothing) = true
+
+function BinDeps.isexecutable(filepath)
+    if isfile(filepath)
+        res = false
+        @unix_only begin
+            res = ccall(:access, Cint, (Ptr{Uint8}, Cint), filepath, 1) == 0
+        end
+        # unfortunately there is no easy way to tell if this process can execute a file on Windows
```
nalimilan: Would there be a way to at least check whether the file is in an executable format?

src/dependencies.jl
```diff
@@ -871,10 +1049,22 @@ macro install(_libmaps...)
...
+                        exec_deps = Dict()
                         for libkey in keys($libmaps)
-                            ((cached = get(load_cache,string(libkey),nothing)) === nothing) && continue
-                            println(depsfile, "@checked_lib ", $libmaps[libkey], " \"", escape_string(cached), "\"")
+                            cached = get(load_cache, string(libkey), nothing)
+                            if cached !== nothing
+                                (dep, path) = cached
+                                if isa(dep, BinDeps.LibraryDependency)
+                                    println(depsfile, "@checked_lib ", $libmaps[libkey], " \"", escape_string(path), "\"")
+                                else
```
nalimilan: Maybe do `elseif isa(dep, BinDeps.ExecutableDependency)` for clarity, and add a final `else` branch raising an error.

nalimilan:
We should really recommend only one method and deprecate the others.

This looks good in general -- at least for what I understand since I don't really know that code.

> Currently install will write a variable _exec to deps.jl which contains a dictionary of dependency names and paths to executables. Please let me know if there is a better way.

What's the reason for the difference with libraries, for which we create separate variables?

src/dependencies.jl
iamed2: The best we could do is ensure the file type ends with one of the PATHEXT extensions, and optionally check files that end in ".exe" for the executable marker (`MZ`). I'll implement that.

src/dependencies.jl
iamed2: Yep

iamed2:
Just realized the call overloading syntax I used has a 0.4 Compat fix but not a 0.3 one. Should I continue to support 0.3?

nalimilan:
I'd say it's fine to drop 0.3 now. Anyway the package hasn't required many fixes in the recent months.

iamed2:
Ok, I'll submit a separate PR for that.

tkelman:
I think this has to be one of the last packages to drop 0.3 support. What's the Compat issue?

iamed2:
Parametric call overloading for types (bug with kw args fixed in https://github.com/JuliaLang/Compat.jl/pull/188), which in v0.5 is just `(::Type{T}){T<:Supertype}(args...) = ...`. As I recall you just couldn't do that in 0.3 and there is no way to fake it such that behaviour is consistent with what is expected.

tkelman:
Can you do `function dependency_constructor(T, context::PackageContext, name; properties...)` ?

iamed2:
That's what I would have done had people wanted to continue 0.3 support.

tkelman:
That didn't come up until 7 hours ago? If it were any package other than BinDeps or Compat I'd be much more in favor of dropping 0.3 support. But here the likelihood of having to backport bugfixes to fix some future issue and keep 0.3 working is somewhat high. Would making this feature support 0.3 be difficult enough to warrant the ongoing maintenance of needing a backport branch here?

iamed2:
Nope, I'll just change it. I guess I didn't communicate very well the first time; I did not mean to imply that it was very difficult to support v0.3, I just wanted to know whether people cared about v0.3 still. Sounds like they do.

tkelman:
A few packages are still actively tagging new versions that support 0.3, e.g. https://github.com/JuliaLang/METADATA.jl/pull/5088 just today. So I think the roots of the dependency tree need to wait until that stops being the case. Or we could press the issue with sufficient warning to the community, if it becomes worth maintaining multiple branches.

iamed2:
New commit supports 0.3 and puts paths into variables rather than a dict. Added a test as well. I did not address the recommended way of loading dependencies, though I think `deps.jl` works better and leaks fewer abstractions.

iamed2:
I've been trying to fix something on Windows for a while and I came across what seems like a bug. Seems like `install` only attempts to `satisfy!` with the platform's defaults, which means `SimpleBuild` will never run on Windows even if it satisfies the dependency.

tkelman:
SimpleBuild has to be added to the defaults to work on Windows. Trying to compile anything on Windows users' systems at install time very rarely works.

ararslan:
Bump. I'm currently in need of this functionality so I'd love to see this go live!

src/dependencies.jl
```diff
@@ -42,9 +53,48 @@ srcdir(dep) = joinpath(depsdir(dep),"src")
...
 
-function _library_dependency(context::PackageContext, name; properties...)
+# normal contents of %PATHEXT% on Windows
+const win_pathext = (".COM", ".EXE", ".BAT", ".CMD", ".VBS", ".VBE", ".JS", ".JSE", ".WSF", ".WSH", ".MSC")
+const win_case_pathext = [collect(win_pathext); map!(lowercase, collect(win_pathext))]
+
+function BinDeps.isexecutable(filepath)
+    if isfile(filepath)
+        res = false
+        @unix_only begin
```
ararslan: Should use Compat and `@static is_*` these calls

iamed2:
I couldn't get things working on all platforms. The testing infrastructure is just not there, and there is functionality partially implemented across many parts of the code, which makes it hard to change. At this point, I am abandoning this PR and waiting for BinDeps to be replaced (I was working on this for Invenia and given the difficulties I've encountered both using and editing BinDeps we've decided not to use it). Feel free to branch off this PR and continue the work.

src/dependencies.jl
```diff
@@ -655,10 +870,10 @@ function can_provide(p::PackageManager,opts,dep)
     end
 end
 
-issatisfied(dep::LibraryDependency) = !isempty(_find_library(dep))
+issatisfied(dep::Dependency) = !isempty(_find_dependency(dep))
 
-allf(deps) = [dep => _find_library(dep) for dep in deps.deps]
-function satisfied_providers(deps::LibraryGroup, allfl = allf(deps))
+allf(deps) = [dep => _find_dependency(dep) for dep in deps.deps]
```
ararslan: This is deprecated syntax but IIRC there's no `@compat` for it. I think you can do
```julia
Dict([Pair(dep, _find_dependency(dep)) for dep in deps.deps])
```
to get around that?

ararslan:
Ah, sad day. Regardless, thanks for your hard work on this! It's certainly inspiring.

Do you have any pointers as to how to declare executable dependencies if not through BinDeps? How is Inventia getting by without it?

iamed2:
Thanks!

We're generally handling it outside of Julia by making containers/instances with what we require already installed, or having separate install scripts. In the case of https://github.com/invenia/FTPClient.jl we just bundled the dependencies in the repo. So we go in both directions from BinDeps; either require it to be installed, or install a specific version ourselves in the build script.

PR #215, opened 2016-05-02
Drop julia 0.3 support in advance of #214

iamed2:
https://github.com/JuliaLang/BinDeps.jl/pull/214#issuecomment-216364063

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.3
   - 0.4
   - nightly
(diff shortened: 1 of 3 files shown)


nalimilan:
It would have been more logical to put this in the same PR as the one actually adding new features. Though now it's done we can merge both at the same time.

iamed2:
I thought this would be easy to get merged, then since it's in master I wouldn't have to reorder my commits on the other branch.

iamed2:
Closing after feedback from Tony

PR #216, opened 2016-05-08
Fix depwarn on 0.5 about symbol and ASCIIString

yuyichao:
Requires a new tag of Compat.jl for `symbol`. I'm not sure if there's a better way to handle `ASCIIString` since there's apparently other packages that depend on this type (it's used in `Conda` for example).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.3
 URIParser
 SHA
-Compat 0.7.1
+Compat 0.7.14+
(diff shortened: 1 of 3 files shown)


nalimilan:
Couldn't you add a type parameter `T<:AbstractString` instead? I don't think we care about the exact type of the string. Same for most cases where `ASCIIString` or `ByteString` appear.

yuyichao:
Whether that is doable depend on the exact/expected usage. Given the low test coverage of this package and the span of supported versions I'd like to change as little as possible on old versions.

nalimilan:
Really, I don't think it could cause any issue. All package names are ASCII, so in practice on 0.4 the field will always end up `ASCIIString`.

yuyichao:
Closed in favor of #217

PR #217, opened 2016-05-11
Fix the String deprecation

timholy:


yuyichao:
Maybe add the symbol fix and close https://github.com/JuliaLang/BinDeps.jl/pull/216

timholy:
Good idea, I hadn't noticed it.

src/dependencies.jl
```diff
@@ -106,7 +106,7 @@ end
 can_use(::Type{AptGet}) = has_apt && OS_NAME == :Linux
 package_available(p::AptGet) = can_use(AptGet) && !isempty(available_versions(p))
 function available_versions(p::AptGet)
-    vers = ASCIIString[]
+    vers = String[]
```
yuyichao: This is potentially breaking if anyone ever used the result....

test/runtests.jl
```diff
@@ -5,8 +5,10 @@ using Compat
 using BinDeps
 
 Pkg.add("Cairo")  # Tests apt-get code paths
+Pkg.build("Cairo")
```
tkelman: doesn't this happen already with `Pkg.add` ?

src/dependencies.jl
tkelman: shall we tag Compat again and use `Compat.ASCIIString` to be on the safe side here then?

tkelman: Actually hold off on tagging Compat again, the way the String deprecation was done broke a lot of existing packages on 0.4 and should be fixed first.

yuyichao: I've just tagged a version locally, I'll wait for now then, especially since `Compat.ASCIIString` and `Compat.UTF8String` works even without JuliaLang/Compat.jl#198

test/runtests.jl
timholy: Not when it's already installed, i.e., if you're running the BinDeps tests locally.

yuyichao:
Amended to use `Compat.ASCIIString[]` in the [one case](https://github.com/JuliaLang/BinDeps.jl/pull/217/files#diff-897099f842b4091bb0bfdae2970b855dR109) that might be breaking.

Will merge later today after the CI finishes.

test/runtests.jl
tkelman: good call. I think we should actually move the `Pkg.add` lines to `test/REQUIRE` instead, so we make slightly fewer permanent changes to the users' package state by running the tests here

PR #221, opened 2016-05-13
BinDeps keyword argument validation

staticfloat:
Tested on OSX.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -29,6 +29,21 @@ type LibraryGroup
     deps::Vector{LibraryDependency}
 end
 
+# List of allowed BinDeps provider keyword options
+const provider_kwopts = [
+    :unpacked_dir,
+    :force_rebuild,
+    :filename,
+    :sha,
+]
+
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks for this! Though looks like you need to support `os` (or at least accept it with a deprecation warning).

staticfloat:
Yep, missed a bunch because I was only looking for `get(opts, :arg)` instead of `haskey(opts, :arg)`.

tkelman:
Bump. I forget, was this ready to go?

staticfloat:
We should run the tests again just in case.

staticfloat:
There, I just rebased onto `master`.  Let's make sure everything is still green.

tkelman:
Apparently HttpParser hasn't been fixed since we removed the 0.4 deprecations on nightly. HttpParser has some API changes not yet resolved that mean it currently can't be tagged until an upgrade path is put in place for existing installs... and windows binaries are rebuilt for the new version. Will need a more precise validation callback etc but this is an issue for a different repo...

PR #222, opened 2016-05-15
fixup after the great String rename

StefanKarpinski:
It looks like @yuyichao already fixed this but this seems like it would also be needed.


src/dependencies.jl
```diff
@@ -106,7 +106,7 @@ end
 can_use(::Type{AptGet}) = has_apt && OS_NAME == :Linux
 package_available(p::AptGet) = can_use(AptGet) && !isempty(available_versions(p))
 function available_versions(p::AptGet)
-    vers = Compat.ASCIIString[]
```
yuyichao: This is intentionally left as `Compat.ASCIIString` so that the code won't break on 0.4.

yuyichao: Ref https://github.com/JuliaLang/Compat.jl/pull/197#discussion_r62842265 and you need https://github.com/JuliaLang/Compat.jl/pull/198 to get rid of the depwarn.

Do you have evidence that the breaking of API will not affect any users of `BinDeps.jl`?

src/dependencies.jl
```diff
@@ -106,7 +106,7 @@ end
...
+    vers = Compat.String[]
```
yuyichao: Also since [`String` is imported](https://github.com/JuliaLang/BinDeps.jl/blob/1e03521858e51ea2fa681fbdf75bd38bbcf1d092/src/BinDeps.jl#L4) this should be `String` and not `Compat.String`.

PR #224, opened 2016-05-22
don't use @ os_only macros

pfitzseb:
also, `OS_NAME` doesn't exist anymore

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -47,7 +47,7 @@ module BinDeps
     function download_cmd(url::AbstractString, filename::AbstractString)
         global downloadcmd
         if downloadcmd === nothing
-            for download_engine in @static is_windows() ? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
+            for download_engine in is_windows() ? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
                 if download_engine == :powershell
                     checkcmd = `$download_engine -NoProfile -Command ""`
                 else
@@ -93,9 +93,7 @@ module BinDeps
             end
             error("I don't know how to unpack $file")
         end
-    end
-
-    @static if is_windows()
+    else
         function unpack_cmd(file,directory,extension,secondary_extension)
             if((extension == ".Z" || extension == ".gz" || extension == ".xz" || extension == ".bz2") &&
                    secondary_extension == ".tar") || extension == ".tgz" || extension == ".tbz"
@@ -299,7 +297,7 @@ module BinDeps
     (|)(a::BuildStep,b::BuildStep) = SynchronousStepCollection()
     function (|)(a::SynchronousStepCollection,b::SynchronousStepCollection)
         if(a.cwd==b.cwd)
-          append!(a.steps,b.steps)
+            append!(a.steps,b.steps)
         else
             push!(a.steps,b)
         end
@@ -372,8 +370,7 @@ module BinDeps
             end
             @dependent_steps ( setenv(cmd, adjust_env(a.env)), )
         end
-    end
-    @static if is_windows()
+    else
         lower(a::MakeTargets,collection) = @dependent_steps ( setenv(`make $(!isempty(a.dir)?"-C "*a.dir:"") $(a.targets)`, adjust_env(a.env)), )
     end
     lower(s::SynchronousStepCollection,collection) = (collection|=s)
@@ -429,9 +426,7 @@ module BinDeps
                     MakeTargets("install";env=env)
                 end
             end
-        end
-
-        @static if is_windows()
+        else
             @dependent_steps begin
                 begin
                     ChangeDirectory(s.src)
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -47,7 +47,7 @@ module BinDeps
     function download_cmd(url::AbstractString, filename::AbstractString)
         global downloadcmd
         if downloadcmd === nothing
-            for download_engine in is_windows() ? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
+            for download_engine in (is_windows() ? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch))
                 if download_engine == :powershell
                     checkcmd = `$download_engine -NoProfile -Command ""`
                 else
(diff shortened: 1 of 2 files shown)


src/BinDeps.jl
```diff
@@ -412,22 +416,32 @@ module BinDeps
...
-    	@windows_only @dependent_steps begin
-    		begin
-                ChangeDirectory(s.src)
-    			@windows_only FileRule(isempty(s.config_status_dir)?"config.status":joinpath(s.config_status_dir,"config.status"),setenv(`sh -c $cmdstring`,env))
-                FileRule(s.libtarget,MakeTargets())
-                MakeTargets("install")
+    	@static if is_windows()
+            @dependent_steps begin
+        		begin
+                    ChangeDirectory(s.src)
```
nalimilan: Wrong indentation here.

nalimilan:
Thanks! Though we'll have to add compatibility macros to Compact.jl before merging this.

src/BinDeps.jl
pfitzseb: Yeah, whitespace was messed up in those files. Thanks for pointing that out!

src/dependencies.jl
```diff
@@ -603,11 +602,11 @@ function check_system_handle!(ret,dep,handle)
 end
 
 # Default installation method
-if OS_NAME == :Darwin
+@static if is_apple()
```
tkelman: shouldn't need to be static

src/BinDeps.jl
```diff
@@ -47,7 +47,7 @@ module BinDeps
     function download_cmd(url::AbstractString, filename::AbstractString)
         global downloadcmd
         if downloadcmd === nothing
-            for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
+            for download_engine in @static is_windows() ? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
```
tkelman: only need to use `@static` for ccall's

src/BinDeps.jl
```diff
@@ -95,7 +95,7 @@ module BinDeps
         end
     end
 
-    @windows_only begin
+    @static if is_windows()
```
tkelman: just `else` on the above

src/BinDeps.jl
```diff
@@ -298,12 +298,12 @@ module BinDeps
...
     (|)(a::BuildStep,b::BuildStep) = SynchronousStepCollection()
     function (|)(a::SynchronousStepCollection,b::SynchronousStepCollection)
-    	if(a.cwd==b.cwd)
-  		append!(a.steps,b.steps)
-    	else
-    		push!(a.steps,b)
-    	end
-    	a
+        if(a.cwd==b.cwd)
+          append!(a.steps,b.steps)
```
tkelman: may as well fix the spacing all the way

src/BinDeps.jl
```diff
@@ -361,27 +361,31 @@ module BinDeps
...
+            end
+            if(!isempty(a.targets))
+                cmd = `$cmd $(a.targets)`
+            end
+            @dependent_steps ( setenv(cmd, adjust_env(a.env)), )
         end
-        @dependent_steps ( setenv(cmd, adjust_env(a.env)), )
     end
-    @windows_only lower(a::MakeTargets,collection) = @dependent_steps ( setenv(`make $(!isempty(a.dir)?"-C "*a.dir:"") $(a.targets)`, adjust_env(a.env)), )
+    @static if is_windows()
```
tkelman: just `else`

src/BinDeps.jl
```diff
@@ -412,47 +416,57 @@ module BinDeps
...
             end
         end
 
-    	@windows_only @dependent_steps begin
-    		begin
-                ChangeDirectory(s.src)
-    			@windows_only FileRule(isempty(s.config_status_dir)?"config.status":joinpath(s.config_status_dir,"config.status"),setenv(`sh -c $cmdstring`,env))
-                FileRule(s.libtarget,MakeTargets())
-                MakeTargets("install")
+        @static if is_windows()
```
tkelman: `else`

pfitzseb:
Alright, I've cleaned this up as per the comments (thanks @tkelman!) and gotten rid of a few more `@static`s -- IIUC, all usages of `@static` at the top level are unnecessary, but inside of method definitions there's a performance win because the `if` doesn't have to be evaluated at runtime, correct?
If so, I think I've used it correctly now.

tkelman:
I just merged `@static` into Compat, it'll need to be tagged though (then that tag's version used in REQUIRE here). And I don't think we've figured out the Compat story for `is_windows` etc yet - I believe those existed in Base but not exported before?

pfitzseb:
[Not really](https://github.com/JuliaLang/julia/commit/694413abca84b1745afe000f05a1676e899e6a43) -- there was only `is_unix`. So we need to add all those function to Compat?

tkelman:
Yeah, you're right, my memory was faulty there. Yes these need to be implemented in Compat.

src/BinDeps.jl
```diff
@@ -47,7 +47,7 @@ module BinDeps
     function download_cmd(url::AbstractString, filename::AbstractString)
         global downloadcmd
         if downloadcmd === nothing
-            for download_engine in @windows? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
+            for download_engine in (is_windows() ? (:powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch))
```
pfitzseb: Without the parentheses around the whole ternary this errors with 
```
ERROR: LoadError: syntax: space before "(" not allowed in "? ("
 in include_from_node1(::String) at .\loading.jl:426
 in macro expansion; at .\none:2 [inlined]
 in anonymous at .\<missing>:?
 in eval(::Module, ::Any) at .\boot.jl:225
 in process_options(::Base.JLOptions) at .\client.jl:243
 in _start() at .\client.jl:322
while loading C:\Users\...\.julia\v0.5\BinDeps\src\BinDeps.jl, in expression starting on line 2
```
but works fine if there's no space... Not quite sure what to make of it.

tkelman: That sounds like it may be a parsing bug? Or depending on what gets tokenized first, `?` might be getting treated as a function.

pfitzseb: The latter, I think:
```
julia> for a in is_windows() ? [1 2] : [3 4]
ERROR: syntax: space before "[" not allowed in "? ["
 in eval(::Module, ::Any) at .\boot.jl:225
 in macro expansion at .\REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at .\event.jl:46

julia> for a in is_windows() ?[1 2] : [3 4]
       println(a)
       end
ERROR: UndefVarError: ? not defined
 in macro expansion; at .\REPL[9]:1 [inlined]
 in anonymous at .\<missing>:?
 in eval(::Module, ::Any) at .\boot.jl:225
 in macro expansion at .\REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at .\event.jl:46

julia> Base.operator_precedence(:in)
6

julia> Base.operator_precedence(:?)
2
```
So this seems like correct, albeit kinda confusing on first glance.

tkelman: maybe in 0.6 some sanity can be restored to `?` now that it isn't being punned on with these `@windows?` etc fake ternary macros

TotalVerb:
:+1: This would fix building binary dependencies on OSX, which is currently broken on nightly because `is_unix` expects `:Darwin` but the deprecated `Base.OS_NAME` is `:OSX`. I think that's the cause of many spurious Travis failures on OSX nightly in downstream packages.

andreasnoack:
Superseded by #229 and #234

PR #227, opened 2016-06-21
Ac/depwarns

abhijithch:
Fixes deprecation warnings.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -298,12 +298,12 @@ module BinDeps
 
     (|)(a::BuildStep,b::BuildStep) = SynchronousStepCollection()
     function (|)(a::SynchronousStepCollection,b::SynchronousStepCollection)
-    	if(a.cwd==b.cwd)
-  		append!(a.steps,b.steps)
-    	else
-    		push!(a.steps,b)
-    	end
-    	a
+      if(a.cwd==b.cwd)
+      append!(a.steps,b.steps)
+      else
+        push!(a.steps,b)
+      end
+      a
     end
     (|)(a::SynchronousStepCollection,b::Function) = (lower(b,a);a)
     (|)(a::SynchronousStepCollection,b) = (lower(b,a);a)
@@ -361,7 +361,8 @@ module BinDeps
         ret
     end
 
-    @static if is_unix() function lower(a::MakeTargets,collection)
+    @static if is_unix()
+      function lower(a::MakeTargets,collection)
             cmd = `make -j8`
             if(!isempty(a.dir))
                 cmd = `$cmd -C $(a.dir)`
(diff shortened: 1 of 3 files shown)


tkelman:
Add the minimum `Compat` version requirement that introduced the features you're using.

src/BinDeps.jl
```diff
@@ -361,26 +361,27 @@ module BinDeps
...
+            cmd = `make -j8`
+            if(!isempty(a.dir))
+                cmd = `$cmd -C $(a.dir)`
+            end
+            if(!isempty(a.targets))
+                cmd = `$cmd $(a.targets)`
+            end
+            @dependent_steps ( setenv(cmd, adjust_env(a.env)), )
+      end
+  end
```
tkelman: be consistent with indentation. use an `else` for the windows part

src/BinDeps.jl
```diff
@@ -361,26 +361,27 @@ module BinDeps
...
     lower(s::SynchronousStepCollection,collection) = (collection|=s)
 
     lower(s) = (c=SynchronousStepCollection();lower(s,c);c)
 
     #run(s::MakeTargets) = run(@make_steps (s,))
 
     function lower(s::AutotoolsDependency,collection)
-    	@windows_only prefix = replace(replace(s.prefix,"\\","/"),"C:/","/c/")
-    	@unix_only prefix = s.prefix
+    	@static is_windows() ? (prefix = replace(replace(s.prefix,"\\","/"),"C:/","/c/")) : nothing
```
tkelman: use `if ... end` instead of `? : nothing`

abhijithch: I went by the documentation [here](https://github.com/JuliaLang/julia/blob/master/base/osutils.jl#L56), is this

src/BinDeps.jl
abhijithch: Thanks, will update this.

src/BinDeps.jl
tkelman: It's more idiomatic to use `if end` for a single condition, the ternary requires you to go looking for a matching `:` and if what follows the `:` is just `nothing` there's no reason to use it.

src/BinDeps.jl
```diff
@@ -298,12 +298,12 @@ module BinDeps
 
     (|)(a::BuildStep,b::BuildStep) = SynchronousStepCollection()
     function (|)(a::SynchronousStepCollection,b::SynchronousStepCollection)
-    	if(a.cwd==b.cwd)
-  		append!(a.steps,b.steps)
-    	else
-    		push!(a.steps,b)
-    	end
-    	a
+      if(a.cwd==b.cwd)
```
tkelman: be consistent, with 4 spaces like the existing code

src/dependencies.jl
```diff
@@ -468,8 +468,8 @@ function generate_steps(dep::LibraryDependency, h::Autotools,  provider_opts)
     env = Dict{String,String}()
     env["PKG_CONFIG_PATH"] = join(opts[:pkg_config_dirs],":")
     delete!(opts,:pkg_config_dirs)
-    @unix_only env["PATH"] = bindir(dep)*":"*ENV["PATH"]
-    @windows_only env["PATH"] = bindir(dep)*";"*ENV["PATH"]
+    @static is_unix() ? (env["PATH"] = bindir(dep)*":"*ENV["PATH"]) : nothing
+    @static is_windows() ? (env["PATH"] = bindir(dep)*";"*ENV["PATH"]) : nothing
```
tkelman: use `if ... end` here as well

src/dependencies.jl
```diff
@@ -603,19 +606,19 @@ function check_system_handle!(ret,dep,handle)
 end
 
 # Default installation method
-if OS_NAME == :Darwin
+if Sys.KERNEL == :Darwin
```
andreasnoack: Use `is_apple` etc here. If you need `KERNEL` then it has to be `Compat.KERNEL`

andreasnoack:
Superseded by #229 and #234

PR #233, opened 2016-07-14
Add test status badges

ararslan:
I'd add the AppVeyor badge too but I don't have the URL for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
 BinDeps.jl
 ==========
 
-[![Build Status](https://travis-ci.org/JuliaLang/BinDeps.jl.svg?branch=master)](https://travis-ci.org/JuliaLang/BinDeps.jl)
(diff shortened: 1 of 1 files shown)


tkelman:
hm, need to fix dict syntax in cairo and tag winrpm for nightly apparently. i think there's a way to construct the appveyor badge link for public repos.

ararslan:
There definitely is for Travis but the AppVeyor URLs have some kind of hash or other garbage in there. Could ask Milan for the link since it runs under his account but it'll be a while.

tkelman:
I think there's a way of doing it that references the github org rather than the appveyor account hash in the badge url (I asked for them to add this capability some time ago). buried in their docs somewhere

ararslan:
Yep, you're absolutely right. Thanks!

ararslan:
Bump

PR #229, opened 2016-06-27
Fix depwarns on julia v0.5

nkottary:
Fix some deprecation warnings on julia v0.5

PR #234, opened 2016-07-15
Adding some missing Compat-ibilities

pabloferz:


src/BinDeps.jl
```diff
@@ -47,8 +47,8 @@ module BinDeps
     function download_cmd(url::AbstractString, filename::AbstractString)
         global downloadcmd
         if downloadcmd === nothing
-            for download_engine in @windows? ("C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell",
-                    :powershell, :curl, :wget, :fetch) : (:curl, :wget, :fetch)
+            for download_engine in (@static is_windows()? ("C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell",
```
kmsquire: `@static` isn't necessary here (and the deprecation message should probably be changed to reflect this).

yuyichao: At least it is not top-level.

tkelman: the deprecation message is accurate because os_only was a macro, but the majority of uses don't actually need to be a macro.

ararslan:
Note that some of this is addressed in #229.

pabloferz:
All right I'll remove the line with the `@static` since that is already in #229 and just will add what is not in there.

andreasnoack:
@pabloferz This one needs a rebase and there are also some `OS_NAME`s left in `dependencies.jl` which should now be `Compat.KERNEL`. We should get this merged soon such that we can avoid more PRs with the same fixes.

tkelman:
Compat.KERNEL should not be used directly for API comparisons, as the changed values there would break packages on mac and windows.

andreasnoack:
Why would it break on Mac and Windows and how should this then be fixed?

tkelman:
it depends on the exact usage, but Compat.KERNEL has a different value than OS_NAME. The OS name is a part of the BinDeps API in several places so you would have to translate to the existing identifiers.

andreasnoack:
Since `OS_NAME` is deprecated, shouldn't this package start using the `KERNEL` values instead? I also only think that Windows is an issue. On my Mac, both `OS_NAME` and `KERNEL` is `:Darwin`.

tkelman:
Sure, but it needs to tranform them, as the new `:NT` value on windows is not used in the BinDeps API.

src/dependencies.jl
```diff
@@ -545,7 +545,7 @@ function _find_library(dep::LibraryDependency; provider = Any)
         end
         @linux_only begin
             soname = ccall(:jl_lookup_soname, Ptr{UInt8}, (Ptr{UInt8}, Csize_t), lib, sizeof(lib))
-            soname != C_NULL && check_path!(ret,dep,bytestring(soname))
+            soname != C_NULL && check_path!(ret,dep,unsafe_string(soname))
```
tkelman: could you rebase? I think this is the only one that hasn't been gotten elsewhere

yuyichao:
Travis OSX seems to be a network issue.

yuyichao:
The Travis 0.5 failure is like a Base julia bug about method cache that is unrelated to this PR. Will merge later.

PR #243, opened 2016-08-31
only modify deps.jl file if it does not yet exist or content changes

tkelman:


tkelman:
@vtjnash is this what you had in mind with your "fix BinDeps" comments elsewhere?

vtjnash:
lgtm

PR #258, opened 2017-01-17
Work around macro needing Compat functions by use of interpolation

musm:
Courtesy  @TotalVerb

I think this is cleaner

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.4
 URIParser
 SHA
-Compat 0.8.4
+Compat 0.9.5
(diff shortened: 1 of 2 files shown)


tkelman:
I disagree, this is more complicated.

musm:
lol okay.

tkelman:
Turns out that without this we get a deprecation from 0cbd186679ea88dfd0635bf98c3a724766903d44 on 0.4, so this is worth doing after all even though the interpolation could be a little confusing.

tkelman:
Thanks! Nightly failures are issues in other test-dep packages

PR #260, opened 2017-01-28
update deprecated Array constructor

musm:


musm:
master failures might be related to https://github.com/JuliaLang/julia/pull/20084 altough I have not actively looked into this

src/BinDeps.jl
```diff
@@ -194,7 +194,7 @@ end
 
 type Choices <: BuildStep
     choices::Vector{Choice}
-    Choices() = new(Array(Choice,0))
+    Choices() = new(Vector{Choice}())
```
musm: `Choice[]` is also possible here and below, not sure which you prefer?

tkelman: there's some discussion that `Vector{T}()` might get deprecated - hasn't happened yet, but evidently the zero arg form isn't loved by all

musm:
I feel like a lot of the inner constructors here are unnecessary

src/dependencies.jl
```diff
@@ -239,7 +239,7 @@ immutable SystemPaths <: DependencyProvider; end
 
 show(io::IO, ::SystemPaths) = print(io,"System Paths")
 
-using URIParser, Compat
```
musm: module already has a `using Compat` so it is not needed to repeat it.

tkelman:
Thanks

PR #265, opened 2017-02-21
Fix unzip usage on bsd

iblis17:
On FreeBSD and NetBSD:
```
$ unzip 
Usage: unzip [-aCcfjLlnopqtuvyZ1] [-d dir] [-x pattern] zipfile
```

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -91,7 +91,11 @@ if is_unix()
         elseif extension == ".tar"
             return (`tar xf $file --directory=$directory`)
         elseif extension == ".zip"
-            return (`unzip -x $file -d $directory`)
+            @static if is_bsd() && !is_apple()
+                return (`unzip -x $file -d $directory $file`)
+            else
+                return (`unzip -x $file -d $directory`)
+            end
         elseif extension == ".gz"
             return pipeline(`mkdir $directory`, `cp $file $directory`, `gzip -d $directory/$file`)
         end


src/BinDeps.jl
```diff
@@ -91,7 +91,11 @@ if is_unix()
         elseif extension == ".tar"
             return (`tar xf $file --directory=$directory`)
         elseif extension == ".zip"
-            return (`unzip -x $file -d $directory`)
+            @static if is_bsd() && !is_apple()
```
tkelman: do we know which convention the apple unzip follows?

iblis17: I borrowed a machine, it's same as Linux
```
$ uname -a
Darwin MBA 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64
$ unzip
UnZip 5.52 of 28 February 2005, by Info-ZIP.  Maintained by C. Spieler.  Send
bug reports using http://www.info-zip.org/zip-bug.html; see README for details.

Usage: unzip [-Z] [-opts[modifiers]] file[.zip] [list] [-x xlist] [-d exdir]
  Default action is to extract files in list, except those in xlist, to exdir;
  file[.zip] may be a wildcard.  -Z => ZipInfo mode ("unzip -Z" for usage).

  -p  extract files to pipe, no messages     -l  list files (short format)
  -f  freshen existing files, create none    -t  test compressed archive data
  -u  update files, create if necessary      -z  display archive comment
  -x  exclude files that follow (in xlist)   -d  extract files into exdir

modifiers:                                   -q  quiet mode (-qq => quieter)
  -n  never overwrite existing files         -a  auto-convert any text files
  -o  overwrite files WITHOUT prompting      -aa treat ALL files as text
  -j  junk paths (do not make directories)   -v  be verbose/print version info
  -C  match filenames case-insensitively     -L  make (some) names lowercase
  -X  restore UID/GID info                   -V  retain VMS version numbers
  -K  keep setuid/setgid/tacky permissions   -M  pipe through "more" pager
Examples (see unzip.txt for more info):
  unzip data1 -x joe   => extract all files except joe from zipfile data1.zip
  unzip -p foo | more  => send contents of foo.zip via pipe into program more
  unzip -fo foo ReadMe => quietly replace existing ReadMe if archive file newer
```

musm: Is the `@static` really needed here?

Issue #199, opened 2016-01-29
Failure to install packages with "sudo: no tty present and no askpass program specified" error


mpinese:
I've run into an odd problem with building packages with apt dependencies, in which the password prompt for the sudo apt-get fails.  At the moment a simple workaround is just installing the packages manually, but I'm posting what I've got in case you guys can see an easy fix.

The problem occurs on Lubuntu 15.10, when running julia 0.4.2 (staticfloat ppa release) from a local LXterminal session.  I've observed it on Cairo.jl ([_jl_libcairo error with Gadfly + Cairo #124](https://github.com/JuliaLang/Cairo.jl/issues/124#issuecomment-176607173) ) and HDF5.jl (see code block following).

```
julia> Pkg.build("HDF5")
INFO: Building Blosc
INFO: Building HDF5
Installing dependency hdf5-tools via `sudo apt-get install hdf5-tools`:
sudo: no tty present and no askpass program specified
================================[ ERROR: HDF5 ]=================================

LoadError: failed process: Process(`sudo apt-get install hdf5-tools`, ProcessExited(1)) [1]
while loading /home/mark/.julia/v0.4/HDF5/deps/build.jl, in expression starting on line 24

================================================================================
```

Hope that's of some use.  Let me know if you need any more logs etc. -- I know Lubuntu isn't a very common distro!  Thanks.

zhenpan:
Hi Mark,

I got the same problem when Pkg.add("HDF5"). Could you please give more details about "how manually the packages" ?  Thank you.
Zhen

yuyichao:
`sudo apt-get install hdf5-tools`

zhenpan:
Yichao, thank you for the quick reply.

Here is the following problem:  I found HDF5 was actually installed along with my anaconda, so I insert the following command in  .juliarc.jl 

**push!(Libdl.LOAD_PATH, "/home/zhenpan/code/anaconda2/pkgs/hdf5-1.8.15.1-2/lib")**

then 

**Pkg.build("HDF5")**

signal (15): Terminated
while loading no file, in expression starting on line 0
syscall at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
uv__epoll_wait at /usr/bin/../lib/x86_64-linux-gnu/julia/libjulia.so (unknown line)
uv__io_poll at /usr/bin/../lib/x86_64-linux-gnu/julia/libjulia.so (unknown line)
[...]

PR #266, opened 2017-02-26
Fix #261 for Arch Linux

### LINKED ISSUE #261: Dependencies for BinDeps
Hi all, I recently ran into an issue where a package would not build properly because Arch linux does not ship with an `unzip` command which is used by `FileUnpacker` if the file contains a `.zip` extension.

```================================================[ ERROR: HttpParser ]=================================================

LoadError: could not spawn `unzip -x /home/hydrotoast/.julia/v0.5/HttpParser/deps/downloads/v2.7.1.zip -d /home/hydrotoast/.julia/v0.5/HttpParser/deps/src`: no such file or directory (ENOENT)
while loading /home/hydrotoast/.julia/v0.5/HttpParser/deps/build.jl, in expression starting on line 71

======================================================================================================================

===================================================[ BUILD ERRORS ]===================================================

WARNING: HttpParser had build errors.

 - packages with build errors remain installed in /home/hydrotoast/.julia/v0.5
 - build the package(s) and all dependencies with `Pkg.build("HttpParser")`
 - build a single package by running its `deps/build.jl` script

======================================================================================================================
```

The solution was to install `unzip` with the native package manager. I also see that `7z` seems to be an implicit dependency for unpacking on Windows.

Could we add these dependencies to the `deps/build.jl` of this package to automatically resolve these dependencies?

americast:
Fixes #261 for Arch Linux.
Adds deps/build.jl

Gramercy.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,0 +1,16 @@
+s=""
+try
+    s = s * chomp(readstring(pipeline(`cat /etc/issue`, stderr=DevNull)))
+end
+try
+    s = s* chomp(readstring(pipeline(`cat /proc/version`, stderr=DevNull)))
+end
+try
+    if(contains(lowercase(s), "arch"))
+        try
+            run(pipeline(`sudo pacman -S unzip`, stdout=DevNull, stderr=DevNull))
+        catch
+            println("Unable to install zip. Please install it manually.")
+        end
+    end
+end


americast:
@yuyichao Incorporated the changes you requested. Also added support for multiple distros.

Gramercy.

tkelman:
shouldn't rely on which being installed, there are environments where it isn't by default

americast:
@tkelman Removed "which".

Gramercy.

tkelman:
won't this be susceptible to https://github.com/JuliaLang/BinDeps.jl/issues/199?

americast:
@tkelman Yes, it will be. But once #199 is solved, this will work fine.
Gramercy.

PR #267, opened 2017-02-27
Improve FreeBSD support

iblis17:
I'm frustrated with `Pkg.add("HDF5")` on FreeBSD...

I'm also curious about what the relationship between `BinDeps` and `Homebrew` is.
Maybe I should create one for FreeBSD?

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -80,7 +80,7 @@ function download_cmd(url::AbstractString, filename::AbstractString)
     end
 end
 
-if is_unix()
+if is_unix() && KERNEL != :FreeBSD
     function unpack_cmd(file,directory,extension,secondary_extension)
         if ((extension == ".gz" || extension == ".Z") && secondary_extension == ".tar") || extension == ".tgz"
             return (`tar xzf $file --directory=$directory`)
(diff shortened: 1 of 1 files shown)


tkelman:
this is getting a bit overcomplicated. a BinDeps provider for the freebsd package manager would be less code.

iblis17:
hmm... so maybe take the first commit only?
The second commit is for changing `make` -> `gmake`, can this also be done in a BinDeps provider?

tkelman:
would setting an environment variable MAKE=gmake help in some cases?

iblis17:
> would setting an environment variable MAKE=gmake help in some cases?

I simply make `gmake` as default on FreeBSD.

Is this PR small enough to merge now?

src/BinDeps.jl
```diff
@@ -91,18 +91,27 @@ if is_unix()
...
     end
 end
 
+if KERNEL == :FreeBSD
+    # The `tar` on FreeBSD can auto-detect the archive format via libarchive.
+    # The supported format can be found in libarchive-formats(5).
+    # For NetBSD and OpenBSD, libarchive is not available.
+    # For macOS, it does. But the previous unpack function works fine already.
+    function unpack_cmd(file, dir, ext, secondary_ext)
+        tar_args = ["--no-same-owner", "--no-same-permissions"]
```
tkelman: are we using these options anywhere else?

src/BinDeps.jl
```diff
@@ -91,18 +91,27 @@ if is_unix()
...
+        return pipeline(
+            `/bin/mkdir -p $dir`,
```
tkelman: bsd tar doesn't create the dir?

tkelman:
sorry this fell off the radar

src/BinDeps.jl
iblis17: Yes, it's here: https://github.com/JuliaLang/BinDeps.jl/pull/267/files#diff-8bd91b8bf2fc0449a96b712e9150aad7R111 .

FYI, https://github.com/freebsd/freebsd-ports/blob/master/Mk/bsd.port.mk#L736-L738

src/BinDeps.jl
iblis17: Yes, `-C/--directory` will try to `cd` as first.
Or, Can I assume the `$dir` exists before this function is invoked?

iblis17:
Bump?

src/BinDeps.jl
```diff
@@ -80,7 +80,7 @@ function download_cmd(url::AbstractString, filename::AbstractString)
     end
 end
 
-if is_unix()
+if is_unix() && KERNEL != :FreeBSD
```
tkelman: this should be Sys.KERNEL

iblis17: :-o https://github.com/JuliaLang/BinDeps.jl/pull/294

Issue #269, opened 2017-03-06
calling the library in validation


lobingera:
the readme says:

> while the handle is a handle that may be passed to ccall or dlsym to check library symbols or the return value of function.

I've been pointed out here https://github.com/JuliaGraphics/Cairo.jl/pull/175#discussion_r104319014 to check library version in build. But 
```
julia> Pkg.build("Cairo")
INFO: Building Cairo
================================[ ERROR: Cairo ]================================

LoadError: TypeError: validate_cairo_version: in ccall: first argument not a pointer or valid constant expression, expected Ptr{T}, got Tuple{Symbol,Ptr{Void}}
while loading /home/lobi/.julia/v0.5/Cairo/deps/build.jl, in expression starting on line 160

================================================================================
...
```
the code to validate is this
```
# check for cairo version 
function validate_cairo_version(name,handle)
    v = ccall((:cairo_version,handle),Int32,())        
    return v > 11200
end
...
   cairo = library_dependency("cairo", aliases = ["libcairo-2", "libcairo","libcairo.so.2", "libcairo2"], depends = [gobject,fontconfig,libpng], group = group, validate = validate_cairo_version)
...
```

full PR here: https://github.com/JuliaGraphics/Cairo.jl/pull/175

nalimilan:
Not sure whether the form you tried is supposed to work, but in StringEncodings I use:
```julia
function validate_iconv(n, h)
    f = Libdl.dlsym_e(h, "iconv_open")
    f == C_NULL && return false

    cd = ccall(f, Ptr{Void}, (Cstring, Cstring), "ASCII", "UTF-8")
...
```

lobingera:
@nalimilan I had a similar idea on the table because i remembered the problems with getting non-constant term in the first argument of ccall, but the documentation mentioned 
> while the handle is a handle that may be passed to ccall 

so i'd thought, give it a try - maybe they have solved this.  -> Just a documentation issue?

nalimilan:
I suspect that's a documentation issue since `?ccall` doesn't say that a library handler can be passed to it.

lobingera:
@Keno You put this into the documentation, any comments?

tkelman:
a handle should work in ccall, either the documentation is missing in ccall or that stopped working at some point

lobingera:
[...]

PR #270, opened 2017-03-11
clarified validation, handle cannot be directly fed into ccall

lobingera:
see #269

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,8 +135,7 @@ cairo = library_dependency("cairo", aliases = ["libcairo-2", "libcairo"], depend
     e.g. whether or not has the correct version. To do so, pass a function that takes 
     (name,handle) as an argument and returns `true` if the library is usable and `false` 
     it not. The `name` argument is either an absolute path or the library name if it is a
-    global system library, while the handle is a handle that may be passed to `ccall` or
-    `dlsym` to check library symbols or the return value of function. Note however that it 
+    global system library, while the handle is a handle that may be passed to  `dlsym` to check library symbols or the return value of function. Note however that it 
     is invalid to store the `handle`. Instead, use the `@load_dependencies` macro (see below).
     Should the validation return false for a library that was installed by a provider, the 
     provider will be instructed to force a rebuild.


ggggggggg:
This is definitely an improvement, but it would be better with an example.

musm:
bump I agree with @ggggggggg but that should not prevent this from getting merged

musm:
@lobingera  looks like this needs a rebase

tkelman:
should rebase without pulling in a bunch of unrelated commits - if your local branch looks okay with only the commits you want on top of master, then you need to force push over the readme-valid branch of your fork

lobingera:
i shouldn't do git work after a concert and before before breakfast...

lobingera:
@musm @tkelman afaics this is now rebased, the failure on nightly is obscure to me

musm:
nice thanks, osx failure I think is unrelated , been seeing that in several places

README.md
```diff
@@ -135,8 +135,7 @@ cairo = library_dependency("cairo", aliases = ["libcairo-2", "libcairo"], depend
     e.g. whether or not has the correct version. To do so, pass a function that takes 
     (name,handle) as an argument and returns `true` if the library is usable and `false` 
     it not. The `name` argument is either an absolute path or the library name if it is a
-    global system library, while the handle is a handle that may be passed to `ccall` or
-    `dlsym` to check library symbols or the return value of function. Note however that it 
+    global system library, while the handle is a handle that may be passed to  `dlsym` to check library symbols or the return value of function. Note however that it
```
nalimilan: Hmm, this line is too long, it should have been split into two parts.

lobingera: Where is the differentiation between OK and too long?

nalimilan: At 92 chars, according to Julia's style guide (in the manual and in CONTRIBUTING.md).

lobingera: I found it in contributing, but rather read it as: for code. But i don't see it in style guide section?

nalimilan: Maybe it could be made clearer. Anyway you can see that the lines around this one respect this rule.

Issue #274, opened 2017-04-07
suppress output from unpack_cmd


musm:
I find it annoying that `unpack_cmd` dumps the extraction info to julia, typically many pages. Would it be possible to pipe this information into a logfile of the extraction folder instead of having it pollute the julia repl and dump pages of extraction details of a dep?

tkelman:
It's possible to suppress output unless there's a problem, see how WinRPM does it. But it is kind of messy to do.

PR #275, opened 2017-04-10
Suppress unpacking output

musm:
close https://github.com/JuliaLang/BinDeps.jl/issues/274

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -83,21 +83,21 @@ end
 if is_unix()
     function unpack_cmd(file,directory,extension,secondary_extension)
         if ((extension == ".gz" || extension == ".Z") && secondary_extension == ".tar") || extension == ".tgz"
-            return pipeline(`tar xzf $file --directory=$directory`, stdout=DevNull)
+            return (`tar xzf $file --directory=$directory`)
         elseif (extension == ".bz2" && secondary_extension == ".tar") || extension == ".tbz"
-            return pipeline(`tar xjf $file --directory=$directory`, stdout=DevNull)
+            return (`tar xjf $file --directory=$directory`)
         elseif extension == ".xz" && secondary_extension == ".tar"
             return pipeline(`unxz -c $file `, `tar xv --directory=$directory`)
         elseif extension == ".tar"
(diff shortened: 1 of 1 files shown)


tkelman:
I don't think this should unconditionally silence all output, you do still want to see if something unexpected happens. The unix unpackers should mostly be well behaved about what they send to stdout vs stderr, but 7z isn't.

musm:
it will still report errors and I think that's the only important thing from a user stand point. The 7z change on windows at least fixes the annoying issue for me.

If you want I can get rid of the other changed on non windows systems

tkelman:
If you get a corrupted download, you want to see the error.

musm:
what I have right will throw errors, e.g.
```julia

julia> run(pipeline(`7z df.s`, stdout=DevNull))
ERROR: failed process: Process(`7z df.s`, ProcessExited(7)) [7]
Stacktrace:
 [1] pipeline_error(::Base.Process) at .\process.jl:687
 [2] run(::Base.CmdRedirect) at .\process.jl:656

# C.F.

julia> run(pipeline(`7z df.s`, stdout=STDOUT))

7-Zip [64] 9.20  Copyright (c) 1999-2010 Igor Pavlov  2010-11-18


Error:
Incorrect command line
ERROR: failed process: Process(`7z df.s`, ProcessExited(7)) [7]
Stacktrace:
 [1] pipeline_error(::Base.Process) at .\process.jl:687
 [2] run(::Base.CmdRedirect) at .\process.jl:656

```

tkelman:
"Incorrect command line" is the important part of that to see what was wrong.

musm:
I can regex capture the error instead

tkelman:
if there's a way of asking for a file list, would be better to filter that out since that's the majority of the issue here

musm:
Searching the web, there does not seem to be such an option. All the recommended options are to search for "evertyhing is ok" and "error"

tkelman:
or look at the exit code, as winrpm does

or convince the 7zip developer to add a quiet flag

musm:
Ok thanks, I'm going to close this and leave the issue open for a better solution.

PR #277, opened 2017-04-12
move complex ldconfig code into dependency build time

vtjnash:
this is in preparation for base Julia to drop this function

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -146,7 +146,7 @@ function available_version(y::Yum)
     found_uname = false
     found_version = false
     for l in eachline(`yum info $(y.package)`)
-        l = chomp(l)
+        VERSION < v"0.6" && (l = chomp(l))
         if !found_uname
             # On 64-bit systems, we may have multiple arches installed
             # this makes sure we get the right one
@@ -208,7 +208,7 @@ function available_version(z::Zypper)
     ENV2 = copy(ENV)
     ENV2["LC_ALL"] = "C"
     for l in eachline(setenv(`zypper info $(z.package)`, ENV2))
-        l = chomp(l)
+        VERSION < v"0.6" && (l = chomp(l))
         if !found_uname
             found_uname = endswith(l, uname)
             continue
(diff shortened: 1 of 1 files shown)


src/dependencies.jl
```diff
@@ -313,6 +313,53 @@ function provides{T}(::Type{T},providers::Dict; opts...)
...
+            m = match(fmt, line)
+            if m !== nothing
+                sonames[m[1]] = m[2]
+            end
+        end
+        have_sonames[] = true
+    end
+    end
+else
+    let fmt = Regex("^\\s:-l(\\S+)\\.\\S+\\s(.+)\$")
```
iblis17: Hi @vtjnash ,
FreeBSD has this pattern
```julia
fmt = r"^\s+\d+:-l(\S+)\.\S+\s(.+)$"
```

iblis17: Here is some sample:
```
└─[iblis@abeing]% ldconfig -r | head
/var/run/ld-elf.so.hints:
        search directories: /lib:/usr/lib:/usr/lib/compat:/usr/local/lib:/usr/local/lib/compat/pkg:/usr/local/lib/fcitx/qt:/usr/local/lib/gcc5:/usr/local/lib/gcc6:/usr/local/lib/graphviz:/usr/local/lib/mysql:/usr/local/lib/nss:/usr/local/lib/opencollada:/usr/local/lib/perl5/5.24/mach/CORE:/usr/local/lib/qt4:/usr/local/lib/samba4:/usr/local/lib/wine:/usr/local/llvm36/lib:/usr/local/llvm37/lib:/usr/local/llvm38/lib:/usr/local/llvm39/lib:/usr/local/mpi/openmpi/lib:/usr/local/mpi/openmpi/lib/openmpi:/usr/local/pypy-5.6/bin
        0:-lalias.7 => /lib/libalias.so.7
        1:-lncursesw.8 => /lib/libncursesw.so.8
        2:-lavl.2 => /lib/libavl.so.2
        3:-lbegemot.4 => /lib/libbegemot.so.4
        4:-lbsdxml.4 => /lib/libbsdxml.so.4
        5:-lc.7 => /lib/libc.so.7
        6:-lcam.6 => /lib/libcam.so.6
        7:-lcrypt.5 => /lib/libcrypt.so.5
```

```julia
julia> f = r"^\s+\d+:-l(\S+)\.\S+ => (.+)$"
r"^\s+\d+:-l(\S+)\.\S+ => (.+)$"

julia> line = "        0:-lalias.7 => /lib/libalias.so.7"
"        0:-lalias.7 => /lib/libalias.so.7"

julia> match(f, line)
RegexMatch("        0:-lalias.7 => /lib/libalias.so.7", 1="alias", 2="/lib/libalias.so.7")
```

vtjnash: How does it distinguish between something like `libLLVM-3.8.so.1` (with digits in the name) and `libperl.so.5.22` (with multiple digits in the version number)?

vtjnash: I just checked - FreeBSD states that it must have exactly one digit following the dot. Thanks for checking up on this! :)

vtjnash:
bump.

tkelman:
How widely have you tested this?

vtjnash:
Ubuntu and FreeBSD. It doesn't have any meaning on Mac OS or Windows. It's also much more extensible than the current code (which is hard coded in C in base, and sometimes fails in our CI tests - although we just hide the error), so it will be much easier to add support for new OS configurations we discover later.

tkelman:
I meant how many packages.

src/dependencies.jl
```diff
@@ -313,6 +313,55 @@ function provides{T}(::Type{T},providers::Dict; opts...)
...
+
+if is_windows() || is_apple()
+    function read_sonames()
+        have_sonames[] = true
+    end
+elseif is_linux()
+    let ldconfig_arch = Dict(:i686 => "x32",
+                             :x86_64 => "x86-64",
+                             :aarch64 => "AArch64"),
+        arch = get(ldconfig_arch, Sys.ARCH, ""),
```
tkelman: Sys.ARCH can end up as i586, possibly i386 or i486 on some distro builds

vtjnash:
Ah, thanks. I thought we normalized all of those, but checking our Make.inc file, I see that we don't.

musm:
bump?

tkelman:
Did you test this against any packages?

tkelman:
I repeat, how many packages did you test this against?

tkelman:
Can you give a specific example where the old code does the wrong thing?

tkelman:
This is causing problems downstream, see https://github.com/JuliaStats/Rmath.jl/issues/38

tkelman:
Without an actual counterexample, I see zero advantage whatsoever to using this version of the code on versions of julia prior to https://github.com/JuliaLang/julia/pull/22886

vtjnash:
Sure. Can you fix the METADATA version bound to require `julia 0.6.1-`

tkelman:
That fix isn't in every `0.6.1-` version and it's not guaranteed to be in 0.6.1 release until it's been demonstrated to not break anything.

You also haven't provided a concrete test case where the old code did the wrong thing, so I'm not convinced this version is ever preferable.

Issue #196, opened 2016-01-08
@load_dependencies() doesn't work as advertised


nalimilan:
README.md says that:
> This will make all your libraries available as variables named by the names you gave the dependency. E.g. if you declared a dependency as
```
    library_dependency("libfoo")
```
The libfoo variable will now contain a reference to that library that may be passed to ccall or similar functions. 

But this doesn't appear to work that way. For example, with the file at https://github.com/nalimilan/iconv.jl/blob/f0ce5cd88921d2aa8b4fabf95a100271872e2e4f/deps/build.jl, if I call `@BinDeps.load_dependencies` from `src/iconv.jl`, I get:
```julie
julia> iconv.libiconv
1-element Array{Any,1}:
 ((System Paths,Dict{Any,Any}()),"/lib64/libc.so.6")
```
which cannot be passed directly to `ccall`.

Actually, all the projects using BinDeps I found load `../deps/deps.jl` directly instead of using `@load_dependencies`. This is no ideal, since they need to replicate the checks about the existence of the file and the error message. And it's particularly bad with regard to the docs, as there is apparently no reference to get this to work. Is this behavior a mere bug?

tkelman:
So I think the reason `@load_dependencies` is effectively deprecated is to avoid a runtime dependency on BinDeps. With package precompilation the load-time cost of that may be low enough to not matter any more. That doesn't necessarily mean `@load_dependencies` works properly though.

nalimilan:
I hadn't thought about that. Maybe saving a few tenths of seconds is worth it, though. I would be fine with recommending `include(join.path(".., "deps", "deps.jl")`, but we should update the docs and deprecate `@load_dependencies`.

tkelman:
It is generally nicer to check for the existence of deps.jl first to give a more specific error message if it wasn't created yet. Maybe the load_dependencies stuff would handle that, if it worked.

nalimilan:
Yes, we should at least give an additional line in the examples which prints an error message when the file doesn't exist.

Gnimuc:
Any progress on this and #100?

PR #282, opened 2017-05-06
update reccomended way to load dependencies

ggggggggg:
Reflecting #196, stop recommending `@load_dependencies` since it doesn't work. Instead provide a few lines of code taken from HDF5.jl that will load deps.jl. [ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ provides(Sources,URI("http://libvirt.org/sources/libvirt-1.1.1-rc2.tar.gz"),libv
 
 To load dependencies without a runtime dependence on BinDeps, place code like the following near the start of the module's primary file. Don't forget to change the error message to reflect the name of your package.
 ```julia
-const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")
+const depsfile = joinpath(dirname(@__FILE__), "deps", "deps.jl")
 if isfile(depsfile)
     include(depsfile)
 else
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ To load dependencies without a runtime dependence on BinDeps, place code like th
 near the start of the Package's primary file. Don't forget to change the error message to 
 reflect the name of the package.
 ```julia
-const depsfile = joinpath(dirname(@__FILE__),"..", "deps", "deps.jl")
+const depsfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
 if isfile(depsfile)
     include(depsfile)
 else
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -254,42 +254,16 @@ provides(Sources,URI("http://libvirt.org/sources/libvirt-1.1.1-rc2.tar.gz"),libv
...
-@BinDeps.load_dependencies
-```
-
-This will make all your libraries available as variables named by the names you gave 
-the dependency. E.g. if you declared a dependency as
-
+To load dependencies without a runtime dependence on BinDeps, place code like the following near the start of the module's primary file. Don't forget to change the error message to reflect the name of your package.
 ```julia
-	library_dependency("libfoo")
+const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")
```
tkelman: doing `dirname(dirname(@__FILE__))` instead of `dirname(@__DIR__)` would avoid a Compat dependency and work on 0.5

tkelman: it is one level different though, so would need another dirname or another `"..",` argument

ggggggggg:
Fixed your comment and added back in the description which variables the dependencies are loaded into.

README.md
```diff
@@ -254,41 +254,25 @@ provides(Sources,URI("http://libvirt.org/sources/libvirt-1.1.1-rc2.tar.gz"),libv
...
-
-if you do not want to change the names of the variables that these libraries get 
-stored in, you may use
-
+To load dependencies without a runtime dependence on BinDeps, place code like the following
+near the start of the Package's primary file. Don't forget to change the error message to 
+reflect the name of the package.
 ```julia
-@BinDeps.load_dependencies [:libfoo=>:_foo, :libbar=>:_bar]
+const depsfile = joinpath(dirname(@__FILE__), "deps", "deps.jl")
```
tkelman: should either call `dirname` twice or do `dirname(@__FILE__), "..", "deps", "deps.jl"`

ggggggggg:
Thanks. I actually tested it this time, and it seems to work.

README.md
```diff
@@ -254,42 +254,26 @@ provides(Sources,URI("http://libvirt.org/sources/libvirt-1.1.1-rc2.tar.gz"),libv
...
-The basic usage is very simple:
-
-```jl
-using BinDeps
-@BinDeps.load_dependencies
+To load dependencies without a runtime dependence on BinDeps, place code like the following
+near the start of the Package's primary file. Don't forget to change the error message to 
+reflect the name of the package.
+```julia
+const depsfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
```
musm: better would be `joinpath(dirname(@__DIR__), "deps", "deps.jl")`   which needs Compat

tkelman: see previous versions, needing Compat makes it not better

musm: I see now. But isn't it best practice to encourage the latest syntax and let Compat take care of backwards compatibility.

tkelman: Only if people want to use Compat, which not everyone does.

README.md
```diff
@@ -254,42 +254,26 @@ provides(Sources,URI("http://libvirt.org/sources/libvirt-1.1.1-rc2.tar.gz"),libv
...
+if isfile(depsfile)
+    include(depsfile)
+else
+    error("HDF5 not properly installed. Please run Pkg.build(\"HDF5\")")
```
musm: `Please run ....  then restart Julia.`

ggggggggg:
updated

musm:
bump

musm:
Thanks!

PR #285, opened 2017-05-11
Remove old compat defs

musm:


src/dependencies.jl
```diff
@@ -884,9 +884,7 @@ macro install(_libmaps...)
                         println(depsfile_buffer)
                         println(depsfile_buffer, "# Load-hooks")
                         println(depsfile_buffer, join(load_hooks,"\n"))
-                        # the following interpolation is needed because take! and String are imported from Compat to support older version of Julia
-                        # interpolated functions are real Julia objects and are accessible from everywhere
-                        depsfile_content = chomp($String($take!(depsfile_buffer)))
+                        depsfile_content = chomp(String(take!(depsfile_buffer)))
```
tkelman: this is needed even on 0.5

musm: Oh right take is 0.6, String is good though

musm:
g2g on my end

musm:
Retrigerring ci although failures are unrelated

musm:
One flaky download failure on windows

musm:
bump

src/dependencies.jl
```diff
@@ -864,15 +864,15 @@ macro install(_libmaps...)
                         depsfile_buffer = IOBuffer()
                         println(depsfile_buffer,
                             """
-                            # This is an auto-generated file; do not edit
+                            # This is an auto-generated file; do not edit this file.
```
tkelman: changing the content here (and below) will force every package to rebuild unnecessarily

musm: I do not believe this is the case.
I checked out master and did `Pkg.update`  then I checked out this PR and did `Pkg.update` and none of the packages rebuilt.

tkelman: not when BinDeps updates, but when they update. see below where the content of deps.jl is compared.

musm: Oh I see, that makes sense. I think it is still worthy to update and remove the dead code here, even if it does force rebuilding the code. I think this is reasonable since it will be triggered upon a an update in the package. And hopefully this will not need to be changed again in the future.

src/dependencies.jl
```diff
@@ -1005,7 +1005,7 @@ macro load_dependencies(args...)
         end
         s = Symbol(sym)
         errorcase = Expr(:block)
-        push!(errorcase.args,:(error("Could not load library "*$(dep.name)*". Try running Pkg.build() to install missing dependencies!")))
+        push!(errorcase.args,:(error("Could not load library "*$(dep.name)*". Try running `Pkg.build(package)` to install missing dependencies!")))
```
musm: updated the text here to match the instructions above regarding `Pkg.build(package)`

PR #286, opened 2017-05-11
Remove old methods for sha_check

musm:


musm:
The travis failures are really odd
```
(new-object net.webclient).DownloadFile( $("http://s3.amazonaws.com/"+$env:JULIAVERSION), "C:\projects\julia-binary.exe")
Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
At line:1 char:44
+ ... .DownloadFile( $("http://s3.amazonaws.com/"+$env:JULIAVERSION), "C:\p ...
+                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
 
Command executed with exception: Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
```

src/dependencies.jl
```diff
@@ -1043,13 +1043,8 @@ using SHA
 function sha_check(path, sha)
     open(path) do f
         calc_sha = sha256(f)
-	# Workaround for SHA.jl API change.  Safe to remove once SHA versions
-	# < v0.2.0 are rare, e.g. when Julia v0.4 is deprecated.
```
tkelman: should set a minimum version for SHA if we do this, but better to leave the flexibility

musm: It won't be a problem since version's less thatn .3 don't have 0.5 support.
https://github.com/staticfloat/SHA.jl/blob/v0.2.2/REQUIRE

tkelman: there's no upper bound so yes they do

musm: Ok so does this repo need 0.3 as a minimum on the `SHA` ?

musm: Or you can just close this PR if you don't like that

tkelman:
I don't see this as necessary

musm:
Ok, It just seems a little odd to incorporate changes in  a package to work around a different packages API changes, especially since 0.4 is now going to be 2 releases old;  but then again, I am not the one that decides these things.

tkelman:
better than breaking or not allowing installation of those versions if the fix is simple

musm:
> better than breaking or not allowing installation of those versions if the fix is simple

True, but as you said adding a minimum would prevent any breakage. *I'm not campaigning for merging this change* but just trying to understand your rationale here.

tkelman:
there were a couple of packages that needed upper bounds for that SHA API change if I remember correctly. would like to avoid holding BinDeps back if those packages are installed, and this doesn't fix any problems

PR #291, opened 2017-05-14
now allows `unpacked_dir="."` syntax for archives without a subdirectory

### LINKED ISSUE #290: `Binaries` provider can't handle archives that unzip to the same directory
I'm downloading a binary `.zip` file for Windows that has the DLL file in the root of the archive, but it doesn't seem that there's an `unpacked_dir` option for that. Currently I have:

```julia
provides(Binaries, URI("http://www.rarewares.org/files/mp3/libmp3lame-3.99.5x64.zip"), lame,
         SHA="97959980490caa1669b1c4af386c3e23fe4ceed0835c04e328c9a5966e4d2521",
         unpacked_dir="")
```

I've also tried `unpacked_dir="."`, but with `""` it interprets it as `libmp3lame-3.99.5x64` (from [this line](https://github.com/JuliaLang/BinDeps.jl/blob/70d4cab063de4f6adbeb63dbf466b4719bc5b767/src/BinDeps.jl#L357)), and with `"."` it complains that "Provider BinDeps.Binaries failed to satisfy dependency libmp3lame" (maybe it's looking for the literal `.` as the filename?

Is there an incantation that can handle this zip file?

ssfrr:
fixes #290

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -374,7 +374,7 @@ function generate_steps(dep::LibraryDependency,h::RemoteBinaries,opts)
     get(opts,:force_rebuild,false) && error("Force rebuild not allowed for binaries. Use a different download location instead.")
     localfile = joinpath(downloadsdir(dep),get(opts,:filename,basename(h.uri.path)))
     # choose the destination to unpack into and the folder/file to validate
-    (dest, target) = if :unpacked_dir in keys(opts)
+    (dest, target) = if haskey(opts, :unpacked_dir)
         if opts[:unpacked_dir] == "."
             # if the archive dumps right in the root dir, create a subdir
             (joinpath(depsdir(dep), dep.name), ".")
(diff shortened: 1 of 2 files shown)


src/dependencies.jl
```diff
@@ -373,10 +373,21 @@ end
 function generate_steps(dep::LibraryDependency,h::RemoteBinaries,opts)
     get(opts,:force_rebuild,false) && error("Force rebuild not allowed for binaries. Use a different download location instead.")
     localfile = joinpath(downloadsdir(dep),get(opts,:filename,basename(h.uri.path)))
+    # choose the destination to unpack into and the folder/file to validate
+    (dest, target) = if :unpacked_dir in keys(opts)
```
tkelman: haskey?

ssfrr: good call. fixed.

tkelman:
not that bad actually. this package is messy and hard to test, and I didn't document this option when I added it 3 years ago, so if you've tested this and it works for you, I think it's okay as is

ssfrr:
Yeah, it works for me locally. Once it's in a tagged version and I tag the package that depends on it (MP3.jl) then I'll be able to come back and add the package as one of the test packages here.

PR #296, opened 2017-05-17
Support the FreeBSD pkg package manager

ararslan:
Works as expected. Unfortunately I had to call the provider BSDPkg to avoid conflict with Julia's own Pkg.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -7,7 +7,8 @@ using Compat
 export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
     Choice, Choices, CCompile, FileDownloader, FileRule,
     ChangeDirectory, FileDownloader, FileUnpacker, prepare_src,
-    autotools_install, CreateDirectory, MakeTargets, SystemLibInstall
+    autotools_install, CreateDirectory, MakeTargets, SystemLibInstall,
+    MAKE_CMD
 
 function find_library(pkg,libname,files)
     Base.warn_once("BinDeps.find_library is deprecated, use Base.find_library instead.")
@@ -523,8 +524,7 @@ function run(s::SynchronousStepCollection)
     end
 end
 
-is_unix() && (make_command = `make -j8`)
-is_windows() && (make_command = `make`)
+const MAKE_CMD = is_bsd() && !is_apple() ? `gmake` : `make`
 
 function prepare_src(depsdir,url, downloaded_file, directory_name)
     local_file = joinpath(joinpath(depsdir,"downloads"),downloaded_file)
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -224,6 +224,48 @@ pkg_name(z::Zypper) = z.package
 
 libdir(z::Zypper,dep) = ["/usr/lib", "/usr/lib32", "/usr/lib64"]
 
+# pkg is the system binary package manager for FreeBSD
+const has_bsdpkg = try success(`pkg -v`) catch e false end
+type BSDPkg <: PackageManager
+    package::AbstractString
+end
+can_use(::Type{BSDPkg}) = has_bsdpkg && Sys.KERNEL === :FreeBSD
+function package_available(p::BSDPkg)
+    can_use(BSDPkg) || return false
+    rgx = Regex(string("^(", p.package, ")(\\s+.+)?\$"))
+    for line in eachline(`pkg search -L name $(p.package)`)
+        ismatch(rgx, line) && return true
+    end
+    return false
+end
+function available_version(p::BSDPkg)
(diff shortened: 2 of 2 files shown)


src/BinDeps.jl
```diff
@@ -523,8 +524,14 @@ function run(s::SynchronousStepCollection)
     end
 end
 
-is_unix() && (make_command = `make -j8`)
-is_windows() && (make_command = `make`)
+const make_command =
```
iblis17: Not sure what is the purpose of `make_command` ?
It does not been exported and not used in the whole codebase either[1].
I guess it's just a stale variable.

[1].https://github.com/JuliaLang/BinDeps.jl/search?utf8=%E2%9C%93&q=make_command&type=

iblis17:
Just FYI: I'm working on this: https://github.com/iblis17/FreeBSDPorts.jl

src/BinDeps.jl
iblis17: and... please also check this: https://github.com/JuliaLang/BinDeps.jl/blob/25c17a227488eb231d9b8bccfee471b469d9c572/src/BinDeps.jl#L390-L398

ararslan: Yes, I saw all of that. I've exported `make_command` as part of this PR so that build steps can use
```
`$make_command`
```
rather than
```
`make`
```

staticfloat: I'm not sure we should be bundling the -j flags in here

ararslan:
> I think this makes sense to live in here

Agreed, particularly since we include the common Linux package managers.

Regarding `make_command`, perhaps it should be in all caps in keeping with the typical style for global constants? It could just be `MAKE`, since that's what should be used in Makefiles to invoke Make.

ararslan:
Failure on 32-bit Windows nightly is unrelated.

src/dependencies.jl
```diff
@@ -224,6 +224,54 @@ pkg_name(z::Zypper) = z.package
...
+            rawversion = chomp(line[findfirst(c->c==':', line)+2:end])
+            dashed = replace(rawversion, "_", "-")
+            if ismatch(Base.VERSION_REGEX, rawversion)
+                return VersionNumber(rawversion)
+            elseif ismatch(Base.VERSION_REGEX, dashed)
+                # Many versions are of the form XX.YY.ZZ_A
+                return VersionNumber(dashed)
+            elseif ismatch(r"^\d+\.\d+\.\d+", rawversion)
+                # Good enough
+                return VersionNumber(matches(r"^(\d+\.\d+\.\d+)", rawversion).captures[1])
```
iblis17: hmm... what is diff between `matches` and `match`

src/dependencies.jl
```diff
@@ -224,6 +224,54 @@ pkg_name(z::Zypper) = z.package
...
+            elseif ismatch(Base.VERSION_REGEX, dashed)
+                # Many versions are of the form XX.YY.ZZ_A
+                return VersionNumber(dashed)
+            elseif ismatch(r"^\d+\.\d+\.\d+", rawversion)
+                # Good enough
+                return VersionNumber(match(r"^(\d+\.\d+\.\d+)", rawversion).captures[1])
+            elseif ismatch(r"^\d", rawversion)
+                # The version number is so messed up, we're basically just gonna guess
+                return VersionNumber(match(r"^(\d)", rawversion).captures[1])
+            else
```
iblis17: Let me explain the freebsd pkg versioning:
`{orig ver}_{PORTREVISION},{PORTEPOCH}`
e.g. openblas -- `0.2.19_1,1`
The `PORTREVISION` is optional. it increase when any changes made in ports tree.
The `PORTEPOCH` is optional as well. It increase when the software author changing the versioning policy. It make pkg do the version comparison correctly. e.g.: libfoo release version `201705` at first, but later libfoo's versioning becomes `1.2`. In order to determine which one is the newer version, we will check the `PORTEPOCH`.

In our case, I purpose that we can simply remove the `_1,1` at first, then do these matching.

ararslan: Right, it's the `{orig ver}` that doesn't seem to have proper restrictions.

iblis17: I provide a counterexample for current matching algo:
`0.35,1` will fail due to the `,1`. If we remove `,1` first, the `Base.VERSION_REGEX` will work fine.

ararslan: Ah, good call, thanks.

src/dependencies.jl
ararslan: The difference is that `match` exists and `matches` doesn't. 😛 Thanks.

ararslan:
32-bit AV failures unrelated (having download issues)

PR #300, opened 2017-05-24
Check for sudo before using it with package managers

ararslan:
This addresses an issue noticed in https://github.com/JuliaLang/julia/pull/21956, where Yum is available so BinDeps tries to use it, but the installation uses `sudo`, which is not installed on the buildbots. With this PR, binary package managers are only usable if `sudo` is available.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -558,4 +558,10 @@ include("show.jl")
 
 @Base.deprecate_binding shlib_ext Libdl.dlext
 
+
+# Definitions that should be evaluated upon initialization, not stored in the cache
+function __init__()
+    global const has_sudo = try success(`sudo -V`) catch err false end
+end
+
 end
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -99,13 +99,11 @@ DEBIAN_VERSION_REGEX = r"^
           ([0-9][a-z0-9.+:~]*))                           # upstream version
 "ix
 
-const has_sudo = try success(`sudo -V`) catch e false end
-
 const has_apt = try success(`apt-get -v`) && success(`apt-cache -v`) catch e false end
 type AptGet <: PackageManager
     package::AbstractString
 end
-can_use(::Type{AptGet}) = has_apt && has_sudo && is_linux()
+can_use(::Type{AptGet}) = has_apt && is_linux()
 package_available(p::AptGet) = can_use(AptGet) && !isempty(available_versions(p))
 function available_versions(p::AptGet)
     vers = Compat.ASCIIString[]
@@ -141,7 +139,7 @@ const has_yum = try success(`yum --version`) catch e false end
 type Yum <: PackageManager
     package::AbstractString
 end
-can_use(::Type{Yum}) = has_yum && has_sudo && is_linux()
+can_use(::Type{Yum}) = has_yum && is_linux()
 package_available(y::Yum) = can_use(Yum) && success(`yum list $(y.package)`)
 function available_version(y::Yum)
     uname = readchomp(`uname -m`)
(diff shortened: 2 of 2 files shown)


staticfloat:
Personally, I don't think we should use this to gate `can_use`; I think we should use this to selectively add in `sudo` to the commands.

ararslan:
@staticfloat Something like this? (See second commit)

ararslan:
Okay, I've now made it so that if the user does not explicitly pass a `sudo=` option, it uses `sudo` if it's found on the system. Does that seem reasonable, @staticfloat?

staticfloat:
Yeah that looks good.  Have you tested this on any platforms?

ararslan:
I don't have Linux handy, and so far no packages are set up to use the `BSDPkg` provider, so I'm not really sure how to test it.

staticfloat:
> I don't have Linux handy

tenor

staticfloat:
Aha!  It took me a while to realize what was going on here.  You need to initialize `has_sudo` within an `__init__()` function, otherwise it remembers whether `sudo` is or is not available from compile-time.

Other than that, this works!

tkelman:
What environments have package managers but not sudo? I'm confused by how that could ever happen

staticfloat:
This is the case with all `docker` images, for example.  Environments in which you are expected to be `root`.

tkelman:
docker suggests that you should run things as an unpriviliged user within the container, and for security reasons we should do so on the buildbots. but guess this should be mostly harmless

staticfloat:
We do run as an unprivileged user on the buildbots, so even with this, the
linked jlbuild invocation won't work.  :P
-E

tkelman:
so this would only help if you happen to be running `Pkg.build` as root, which is probably a bad idea most of the time

ararslan:
I don't think I understand why this wouldn't address the issue in the linked invocation of jlbuild

ararslan:
Is this even useful then?

staticfloat:
Yes, I still think this is a good idea.

tkelman:
> add passwordless sudo

would be better to only allow package installation privileges for the buildbot non-root user, if that's doable

ararslan:
Bump

src/BinDeps.jl
```diff
@@ -558,4 +558,10 @@ include("show.jl")
 
 @Base.deprecate_binding shlib_ext Libdl.dlext
 
+
+# Definitions that should be evaluated upon initialization, not stored in the cache
+function __init__()
+    global const has_sudo = try success(`sudo -V`) catch err false end
```
musm: This shouldn't be  a `global const`  which can cause issues when used in __init__ for precompilation

src/dependencies.jl
```diff
@@ -356,6 +356,7 @@ function provides{T}(::Type{T},providers::Dict; opts...)
     end
 end
 
+sudoname(c::Cmd) = c == `` ? "" : "sudo "
```
musm: I don't understand why include the  space after `sudo` ?

src/BinDeps.jl
ararslan: Shouldn't matter in this case.

src/dependencies.jl
ararslan: So it looks nicer when it's interpolated below.

src/BinDeps.jl
musm: global const is incompatible with precompilation

tkelman: https://github.com/JuliaLang/julia/issues/13817#issuecomment-152157853

src/dependencies.jl
musm: how so ? looks like it could be confusing, I wouldn't expect the sudo name to include a space after

src/BinDeps.jl
ararslan: It's not, plus it doesn't matter in this case

src/dependencies.jl
tkelman: look at the usages

musm: can't you just eliminate this function and use `string` which is shorter

PR #303, opened 2017-06-06
Use a Ref for has_sudo rather than global in __init__

ararslan:
appease the precompile gods

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -558,10 +558,9 @@ include("show.jl")
 
 @Base.deprecate_binding shlib_ext Libdl.dlext
 
-
-# Definitions that should be evaluated upon initialization, not stored in the cache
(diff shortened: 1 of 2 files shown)


src/BinDeps.jl
```diff
@@ -558,10 +558,9 @@ include("show.jl")
 
 @Base.deprecate_binding shlib_ext Libdl.dlext
 
-
-# Definitions that should be evaluated upon initialization, not stored in the cache
+const has_sudo = Ref{Bool}(false)
 function __init__()
-    global const has_sudo = try success(`sudo -V`) catch err false end
+    has_sudo[] = try success(`sudo -V`) catch err false end
```
musm: since you're not using `err` the following would be better `try success(`sudo -V`) catch false end`

Issue #55, opened 2013-09-17
BuildProcess providers on Windows?


carlobaldassi:
I now have a working GLPK Windows installer in an experimental branch of GLPK.jl:

https://github.com/carlobaldassi/GLPK.jl/blob/309ab10f566239f77acedb3d179b518ec741b14c/deps/build.jl

However, I had to somehow force BinDeps to accept `BuildProcess` as a provider on Windows:

```
@windows_only push!(BinDeps.defaults, BuildProcess)

@BinDeps.install

@windows_only pop!(BinDeps.defaults)
```

The issue is that the compiled binaries come in a peculiar directory structure, so I have to manipulate them to install the library under `usr/lib`, and I couldn't find a way to do this with a `Binary` provider.
Is this acceptable (i.e. could something go wrong with the above solution)? Is there a better way to do this?

musm:
bump
I don't understand why we have 
`push!(BinDeps.defaults, BuildProcess)`
on windows?

musm:
I opened a pr to change the default, however seems unlikely to be accepted, so the manual force seems like the only solution in this case. A cleaner solution should be possible either by including it as a default or via  potentially another idea

PR #305, opened 2017-06-16
Add BuildProcess to windows providers

musm:
requested addition https://github.com/JuliaLang/BinDeps.jl/issues/55

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -727,7 +727,7 @@ if is_apple()
 elseif is_linux() || (is_bsd() && !is_apple())
     defaults = [PackageManager,SystemPaths,BuildProcess]
 elseif is_windows()
-    defaults = [Binaries,PackageManager,SystemPaths]
+    defaults = [Binaries,PackageManager,SystemPaths,BuildProcess]
 else
     defaults = [SystemPaths,BuildProcess]
 end


tkelman:
This is unlikely to work for most of these that get written, given the usual lack of development tools on windows.

musm:
What I see in practice is custom  unpackaging and installation scripts that are used within in a 
 `is_windows()` branch that end up using  `BuildProcess` , which makes this patch useful

musm:
I  several others have made the same observations. Would be nice to just have it as a default. Instead of having to resort to, which is what is currently necessary,
```
is_windows() && push!(BinDeps.defaults, BuildProcess)
...
is_windows() && pop!(BinDeps.defaults)
```

musm:
As a note the 32-bit cairo failures are unrelated  see https://github.com/JuliaGraphics/Cairo.jl/issues/192

tkelman:
The problem is the default BuildProcess doesn't have an OS conditional. So this would enable all the BuildProcess providers out there that have been written using configure and make that won't work at all. At least having to go to the extra length of adding it manually is a way of certifying yes, you know it should work, as opposed to default assuming it will work cross-platform which will often be a wrong assumption.

musm:
Closing as it seems that you do not want this

PR #307, opened 2017-06-19
Update debug_context include method (do not merge)

musm:
Will need a Compat to update to include this functionality first


musm:
Compat doesn't typically support unexported functions so this is unlikely to work without custom versioning

PR #309, opened 2017-06-23
Removed trailing whitespace and fixed spelling in README

ninjin:


ninjin:
Call me an optimist, but I doubt that the CI failures are due to README changes.

PR #314, opened 2017-07-20
Drop support for 0.5, fix all 0.7 depwarns

ararslan:
Going forward if there's anything that needs fixing that applies to old versions we can always backport the change and make another 0.5-compatible release. But it makes sense to me to focus on what's easier to maintain going forward. Plus there are a lot of deprecation warnings on nightly which contribute to the thousands of lines of depwarns in Nanosoldier's logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 notifications:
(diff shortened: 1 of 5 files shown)


tkelman:
This is going to actively need to support 0.5 for quite some time.

ararslan:
:( Could use separate branches then I guess? Or just close this.

tkelman:
we could do the other way around I suppose and make an 0.7-only branch for a while

PR #324, opened 2017-07-31
make DEBIAN_VERSION_REGEX a const

musm:


musm:
Wasn't aware of BinDeps2.jl prolly not worth these minor fixes

musm:
Very minor but, this should be a const for perf and safety  cc @ararslan @tkelman

PR #327, opened 2017-09-22
Revert "Add Binaries installation method on Linux"

tkelman:
Reverts JuliaLang/BinDeps.jl#326. Non-glibc linux distros exist and Julia runs on them, glibc-built binaries should never be used there.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -725,7 +725,7 @@ end
 if Compat.Sys.isapple()
     defaults = [Binaries,PackageManager,SystemPaths,BuildProcess]
 elseif Compat.Sys.islinux() || (Compat.Sys.isbsd() && !Compat.Sys.isapple())
-    defaults = [PackageManager,SystemPaths,Binaries,BuildProcess]
+    defaults = [PackageManager,SystemPaths,BuildProcess]
 elseif Compat.Sys.iswindows()
     defaults = [Binaries,PackageManager,SystemPaths]
 else


staticfloat:
I think you're being overruled here Tony, as Alex, Andreas and I all seem to agree that this change is desirable.  Although I sympathize and agree that we should attempt to be as compatible and forward-thinking as we can, I do think that the principle of getting things working as well as possible for the large majority of users first, and fixing corner cases later, trumps that ideal.  Non-glibc linux distros are, for the moment at least, rather rare (especially among Julia users), and perhaps that will change in the next few years.  But we have many more pressing problems than dealing with such compatibility issues, and as I'm sure you can agree,weI only have so much time and energy to spare.

I think it is reasonable to move forward with this incorrect, but much more useful behavior, than to continue to stick with a crippled Linux BinDeps.  I think non-glibc support is just going to have to be driven by user complaints or a brave dev that will work in the proper platform support, because as it stands I don't think we can even tell from within Julia whether it is running on `glibc` or not.

I move to close this as "won't fix right now".

tkelman:
You look at the triple from Sys.MACHINE as I've been fruitlessly encouraging you to do elsewhere. If you're insistent on treating all distros exactly alike, despite a fair amount of work having gone into Julia itself to make it work on Alpine, you can go ahead and pretend that is correct when it isn't. You wouldn't allow Linux binaries to be installed on FreeBSD would you? This is as incorrect as doing that.

The correct way of doing this is to whitelist a set of known glibc triples where this can be enabled. For now I don't think convenience calls for having broken default behavior, even on distros that aren't as widely used.

ararslan:
As of #329, we have an easy way to check whether the system uses glibc. What about something like
```diff
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -725,7 +725,11 @@ end
 if Compat.Sys.isapple()
     defaults = [Binaries,PackageManager,SystemPaths,BuildProcess]
 elseif Compat.Sys.islinux() || (Compat.Sys.isbsd() && !Compat.Sys.isapple())
-    defaults = [PackageManager,SystemPaths,Binaries,BuildProcess]
+    if glibc_version() !== nothing
+        defaults = [PackageManager,SystemPaths,Binaries,BuildProcess]
+    else
+        defaults = [PackageManager,SystemPaths,BuildProcess]
+    end
 elseif Compat.Sys.iswindows()
     defaults = [Binaries,PackageManager,SystemPaths]
 else
```
?

PR #329, opened 2017-10-03
Add glibc_version for checking the glibc version

ararslan:
This PR adds a function I've called `glibc_version` that returns the version of glibc in use by Julia (as accessible from `dlopen`) on glibc-based Linux systems. For all other systems, including musl Linux, this returns `nothing`.

When binaries are available on Linux, this should make it easy to check whether the current system is compatible with the binaries (at least in the sense of a compatible libc).

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -8,7 +8,7 @@ export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
     Choice, Choices, CCompile, FileDownloader, FileRule,
     ChangeDirectory, FileDownloader, FileUnpacker, prepare_src,
     autotools_install, CreateDirectory, MakeTargets, SystemLibInstall,
-    MAKE_CMD
+    MAKE_CMD, glibc_version
 
 function find_library(pkg,libname,files)
     Base.warn_once("BinDeps.find_library is deprecated; use Base.find_library instead.")
@@ -559,6 +559,21 @@ function eval_anon_module(context, file)
         eval(m, body)
     end
     return
+end
+
+"""
+    glibc_version()
+
+For Linux-based systems, return the version of glibc in use. For non-glibc Linux and
+other platforms, returns `nothing`.
+"""
+function glibc_version()
+    Compat.Sys.islinux() || return
+    libc = ccall(:jl_dlopen, Ptr{Void}, (Ptr{Void}, UInt32), C_NULL, 0)
+    ptr = Libdl.dlsym_e(libc, :gnu_get_libc_version)
+    ptr == C_NULL && return # non-glibc
+    v = unsafe_string(ccall(ptr, Ptr{UInt8}, ()))
+    ismatch(Base.VERSION_REGEX, v) ? VersionNumber(v) : nothing
 end
 
 include("dependencies.jl")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,3 +14,18 @@ using GSL
 
 # PR 271
 BinDeps.debug("Cairo")
+
+let gv = glibc_version()
+    if Compat.Sys.islinux()
+        lddv = lowercase(readchomp(`ldd --version`))
+        if contains(lddv, "gnu") || contains(lddv, "glibc")
+            @test isa(gv, VersionNumber)
+            @test gv >= v"1.0.0"
+        else
+            # Assume non-glibc
+            @test gv === nothing
+        end
+    else
+        @test gv === nothing
+    end
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,10 @@
-using Base.Test
 using Compat
 using BinDeps
+if VERSION < v"0.7.0-DEV.2004"
+    using Base.Test
+else
+    using Test
+end
 
 Pkg.build("Cairo")  # Tests apt-get code paths
 using Cairo


src/BinDeps.jl
```diff
@@ -561,6 +561,20 @@ function eval_anon_module(context, file)
...
 
+"""
+    glibc_version()
+
+For Linux-based systems, return the version of glibc in use. For non-glibc Linux and
+other platforms, returns `nothing`.
+"""
+function glibc_version()
+    Compat.Sys.islinux() || return
+    ptr = Libdl.dlsym_e(Libdl.dlopen_e("libc"), :gnu_get_libc_version)
```
yuyichao: `""` should be enough and will avoid dealing with sonames.
Also, you should either use `Libdl.dlopen`, or use `Libdl.dlopen_e` with check on return value before passing it to `dlsym`.

ararslan: What do you mean "`""` should be enough"? Good call regarding checking the return value of `dlopen(_e)`.

yuyichao: Replace the only string in this line with `""`.

ararslan: I'm confused, that doesn't seem to do the same thing:
```
julia> Libdl.dlopen_e("")
Ptr{Void} @0x0000000000000000

julia> Libdl.dlopen("libc")
Ptr{Void} @0x0000000112fafc90
```

test/runtests.jl
```diff
@@ -14,3 +14,17 @@ using GSL
 
 # PR 271
 BinDeps.debug("Cairo")
+
+let gv = glibc_version()
+    if Compat.Sys.islinux()
+        if contains(readchomp(`ldd --version`), "GNU libc")
```
ararslan: This is not a good check; it's failing on Travis

src/BinDeps.jl
yuyichao: This behavior might be version dependent so the more reliable one would be to use `ccall(:jl_dlopen, Ptr{Void}, (Ptr{Void}, UInt32), C_NULL, 0)`.

ararslan: Which behavior is version-dependent? `dlopen_e("")` returns a null pointer and `dlopen_e("libc")` seems to work as expected on all of 0.3-0.7.

yuyichao: Version dependent as in libc version dependent. Not julia version dependent. `libc` is not guarantee to exist.

ararslan: Oh! Weird, I didn't realize that libc could be absent altogether.

yuyichao: `libc.so` does not exist on most systems.

yuyichao: And even when it does, it's basically never the one you want.

ararslan: > on most systems

On most Linux systems? This code only gets run on Linux since the other platforms return early.

yuyichao: Correct. It's a linker script for all glibc.

ararslan:
Comments addressed. How does this look now? The 32-bit Windows failures are unrelated—something to do with WinRPM.

src/BinDeps.jl
```diff
@@ -561,6 +561,21 @@ function eval_anon_module(context, file)
...
 
+"""
+    glibc_version()
+
+For Linux-based systems, return the version of glibc in use. For non-glibc Linux and
+other platforms, returns `nothing`.
+"""
+function glibc_version()
+    Compat.Sys.islinux() || return
+    libc = ccall(:jl_dlopen, Ptr{Void}, (Ptr{Void}, UInt32), C_NULL, 0)
```
staticfloat: Ah, so passing `C_NULL` to `Libdl.dlopen()` doesn't work anymore, but we can still `ccall(:jl_dlopen)` with `C_NULL` to get this behavior?  This seems kind of underhanded.  Is this a sign that we should have a `libc_name` in `Base` a la `libblas_name`?

src/BinDeps.jl
```diff
@@ -561,6 +561,21 @@ function eval_anon_module(context, file)
...
+    ptr = Libdl.dlsym_e(libc, :gnu_get_libc_version)
+    ptr == C_NULL && return # non-glibc
```
staticfloat: Personally, I don't like the `... || return` and `... && return` style, but that's a total style thing and not something I'm asking you to change.  Just whining.  :)

src/BinDeps.jl
```diff
@@ -561,6 +561,21 @@ function eval_anon_module(context, file)
...
+    v = unsafe_string(ccall(ptr, Ptr{UInt8}, ()))
+    ismatch(Base.VERSION_REGEX, v) ? VersionNumber(v) : nothing
```
staticfloat: I wonder if we should print out a warning here?  What should we do if we encounter a `glibc` that we can't get the version of?

src/BinDeps.jl
ararslan: Perhaps worth opening an issue about. Seems useful.

src/BinDeps.jl
ararslan: That's a good question. Doing it this way was more of a precaution; I think `gnu_get_libc_version` is always supposed to produce a valid version number, so maybe an error would be best?

src/BinDeps.jl
yuyichao: i don't think it was ever allowed but I do think it's useful.

src/BinDeps.jl
```diff
@@ -561,6 +561,21 @@ function eval_anon_module(context, file)
...
+"""
+    glibc_version()
+
+For Linux-based systems, return the version of glibc in use. For non-glibc Linux and
+other platforms, returns `nothing`.
+"""
+function glibc_version()
+    Compat.Sys.islinux() || return
+    libc = ccall(:jl_dlopen, Ptr{Void}, (Ptr{Void}, UInt32), C_NULL, 0)
+    ptr = Libdl.dlsym_e(libc, :gnu_get_libc_version)
```
yuyichao: Ideally (pedantically?) you should check if `libc` is `NULL` but this is posix behavior so it's probably find...

ararslan:
```
julia> Libdl.dlopen_e("libc")
Ptr{Void} @0x0000000110471c90

julia> ccall(:jl_dlopen, Ptr{Void}, (Ptr{Void}, UInt32), C_NULL, 0)
Ptr{Void} @0xfffffffffffffffe
```
Should these be the same? If not, the `libc` variable name is probably inaccurate, right?

yuyichao:
> Should these be the same?

No

> If not, the libc variable name is probably inaccurate, right?

Both will give you the right symbol address.

ararslan:
Okay cool, thanks.

src/BinDeps.jl
ararslan: What do you think of this, @yuyichao?

PR #331, opened 2017-10-05
add some help strings to BinDeps

simonbyrne:
also remove a _very_ old deprecated function, and slightly rearrange.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -10,30 +10,10 @@ export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
     autotools_install, CreateDirectory, MakeTargets, SystemLibInstall,
     MAKE_CMD
 
-function find_library(pkg,libname,files)
-    Base.warn_once("BinDeps.find_library is deprecated; use Base.find_library instead.")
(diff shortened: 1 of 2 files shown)


src/BinDeps.jl
```diff
@@ -135,39 +33,79 @@ end
...
+Specifies that the working directory should be changed to `dir`.
+"""
 type ChangeDirectory <: BuildStep
     dir::AbstractString
 end
 
+"""
+    CreateDirectory(dir, mayexist=true) <: BuildStep
+
+Specifies that the directory `dir` should be created. `mayexist` specifies if whether or not the directory may exist.
```
ararslan: "If whether or not" is a little awkward.

src/BinDeps.jl
```diff
@@ -232,6 +177,11 @@ function run(c::Choices)
     end
 end
 
+"""
+    CCompile(src, dest, options, libs) <: BuildStep
+
+Compile C file `src` to `dest`.
```
ararslan: Might be good to provide equivalent pseudocode like `cc src -o dest <options>`, etc.

PR #332, opened 2017-10-06
use jl_read_sonames and jl_lookup_soname on older julia versions

tkelman:
use the code path from before #277 on versions of julia where the new code freezes
ref https://github.com/JuliaStats/Rmath.jl/issues/38

This is needed before master is safe to make another tag from, without breaking on release versions of Julia.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -426,7 +426,7 @@ if VERSION >= v"0.7.0-DEV.1287" # only use this where julia issue #22832 is fixe
     end
 else
     function lookup_soname(lib)
-        if is_linux() || (is_bsd() && !is_apple())
+        if Compat.Sys.islinux() || (Compat.Sys.isbsd() && !Compat.Sys.isapple())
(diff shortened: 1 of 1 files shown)


src/dependencies.jl
```diff
@@ -410,10 +418,20 @@ else
...
-    return get(sonames, s, "")
+if VERSION >= v"0.7.0-DEV.1287" # only use this where julia issue #22832 is fixed
+    lookup_soname(s) = lookup_soname(String(s))
+    function lookup_soname(s::String)
+        have_sonames[] || read_sonames()
+        return get(sonames, s, "")
+    end
+else
+    function lookup_soname(lib)
+        if is_linux() || (is_bsd() && !is_apple())
```
ararslan: Should be `Compat.Sys.islinux() || (Compat.Sys.isbsd() && !Compat.Sys.isapple())`, or stick a `using Compat.Sys: islinux, isapple, isbsd` at the top and drop the `Compat.Sys`s here.

musm:
should this be merged and then tagged?

ararslan:
Need more stuff before tagging.

PR #333, opened 2017-10-07
Disallow binaries by default on non-glibc Linux

ararslan:
Replaces #327

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -722,14 +722,16 @@ function check_system_handle!(ret,dep,handle)
 end
 
 # Default installation method
-if Compat.Sys.isapple()
-    defaults = [Binaries,PackageManager,SystemPaths,BuildProcess]
-elseif Compat.Sys.islinux() || (Compat.Sys.isbsd() && !Compat.Sys.isapple())
-    defaults = [PackageManager,SystemPaths,Binaries,BuildProcess]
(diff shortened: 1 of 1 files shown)


tkelman:
There should really be a way of detecting or annotating an actual minimum here for any given binary.

ararslan:
That would be neat but that sounds like a pretty significant change, since you'd also have to inspect all downloaded binaries and reliably determine the glibc version they require. Worth thinking about, though that's more than I wanted to get into with this PR.

src/dependencies.jl
```diff
@@ -722,14 +722,16 @@ function check_system_handle!(ret,dep,handle)
 end
 
 # Default installation method
-if Compat.Sys.isapple()
-    defaults = [Binaries,PackageManager,SystemPaths,BuildProcess]
-elseif Compat.Sys.islinux() || (Compat.Sys.isbsd() && !Compat.Sys.isapple())
-    defaults = [PackageManager,SystemPaths,Binaries,BuildProcess]
+defaults = if Compat.Sys.isbsd()
+    [Binaries, PackageManager, SystemPaths, BuildProcess]
```
tkelman: I don't think Binaries are safe to enable for non-Apple bsd here.

ararslan: Fixed in https://github.com/JuliaLang/BinDeps.jl/pull/334

Issue #328, opened 2017-09-23
BinDeps Installation requires root


montyvesselinov:
How to avoid installation to attempt `apt-get` but go straight to downloading and compilation from the source.

PR #334, opened 2017-10-07
RFC: Customizable priorities for source building

ararslan:
This introduces two new environment variables: `JULIA_BINDEPS_USE_SUDO` and `JULIA_BINDEPS_BUILD_SOURCE`. The former can be used to avoid checking `sudo -V` and automatically enable or disable `sudo`. The latter allows the user to set their preference for how/when to build from source. See the additions to the README for a more thorough explanation.

`JULIA_BINDEPS_BUILD_SOURCE` provides a solution to #328, and `JULIA_BINDEPS_USE_SUDO` is essentially #325.

### DIFF
--- a/src/BinDeps.jl
+++ b/src/BinDeps.jl
@@ -586,8 +586,27 @@ include("show.jl")
 @Base.deprecate_binding shlib_ext Libdl.dlext
 
 const has_sudo = Ref{Bool}(false)
+const build_source = Ref{String}("default")
+
 function __init__()
-    has_sudo[] = try success(`sudo -V`) catch err false end
+    # install from package managers with sudo
+    has_sudo[] = if haskey(ENV, "JULIA_BINDEPS_USE_SUDO")
+        ENV["JULIA_BINDEPS_USE_SUDO"] == "true"
+    else
+        try
+            success(`sudo -V`)
+        catch err
+            false
+        end
+    end
+
+    # preference for building from source
+    build_source[] = let opt = get(ENV, "JULIA_BINDEPS_BUILD_SOURCE", "default")
+        if !in(opt, ["default", "always", "never", "prefer"])
+            error("Unrecognized value '$opt' for environment variable JULIA_BINDEPS_BUILD_SOURCE")
+        end
+        opt
+    end
 end
 
 end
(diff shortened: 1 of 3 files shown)


ararslan:
Bump @yuyichao, @tkelman, @staticfloat. Would be nice if someone could review this.

src/BinDeps.jl
```diff
@@ -574,18 +574,36 @@ function glibc_version()
...
+const has_sudo = Ref{Bool}(false)
+const build_source = Ref{String}("default")
+
+function __init__()
+    # install from package managers with sudo
+    has_sudo[] = if haskey(ENV, "JULIA_BINDEPS_USE_SUDO")
+        ENV["JULIA_BINDEPS_USE_SUDO"] == "true"
+    else
+        try
+            success(`sudo -V`)
```
bjarthur: can you build a new system image with this line as is?  i can not.  like so:

```
$ cat ~/userimg.jl 
using BinDeps

julia> include(joinpath(JULIA_HOME, Base.DATAROOTDIR, "julia", "build_sysimg.jl"))
julia> build_sysimg("/Users/arthurb/sys.ji", "native", "/Users/arthurb/userimg.jl")

$ julia -J /Users/arthurb/sys.ji.dylib

signal (11): Segmentation fault: 11
while loading no file, in expression starting on line 0
Segmentation fault: 11
```

works though if i simply replace this line with `true`

Pbellive:
As a user who ran into #328 today, I'd like to second the request that the functionality implemented in this PR be added to BinDeps. As far as I can tell, when sudo and a package manager are available on a system but the user does not have permission to run
```
<package manager> install <dependency>
```
`Pkg.build(<package>)` throws an error and exits even if it might be possible to build `<dependency>` from source. The only workaround to install `<dependency>` seems to be to comment out the
```
provides(<package manager>,...
```
lines in relevant build.jl script and re-run `Pkg.build` to trigger building the dependency from source. This isn't a big deal for experienced users but causes difficulty for novice users trying to install Julia packages with binary dependencies and makes the package dirty.

I can't speak to whether or not this PR is the best way to address this issue but I wanted to add a little context about the usefulness of the functionality it provides. Thanks to all the BinDeps devs for their work on BinDeps and to @ararslan for the work on this PR!

PR #319, opened 2017-07-26
fix depwarns under 0.7 and bump REQUIRE and CI to 0.6

Sacha0:
Best! :)

PR #336, opened 2017-10-15
upgrade to 0.6

KristofferC:


src/BinDeps.jl
```diff
@@ -6,8 +6,8 @@ using Compat
 
 export @make_run, @build_steps, find_library, download_cmd, unpack_cmd,
     Choice, Choices, CCompile, FileDownloader, FileRule,
-    ChangeDirectory, FileDownloader, FileUnpacker, prepare_src,
```
KristofferC: The removed ones does not exist in the package.

src/BinDeps.jl
```diff
@@ -297,8 +297,6 @@ macro dependent_steps(steps)
     blk
 end
 
-mypwd() = chomp(readall(`pwd`))
```
KristofferC: Not used, and `readall` doesn't exist.

KristofferC:
bump @ararslan

ararslan:
Sorry, I didn't see this. I guess I'm not following BinDeps on GitHub.

See #319 and #314. As much as I'd like to, I guess we can't drop 0.5 yet. Tony mentioned that we may be able to do 0.7-compatible releases from a separate branch.

src/BinDeps.jl
ararslan: Hah, that's also equivalent to the Julia function `pwd()`.

src/BinDeps.jl
ararslan: ?? I'm pretty sure `ChangeDirectory` does, as I think I have a package that uses it.

KristofferC:
Just do backports to a 0.5 branch if needed? That's how Documenter does it.

ararslan:
Yes, that's what I'd prefer to do. Any objections, @tkelman?

ararslan:
CI failures are unrelated. Travis on Linux 0.7 is an error fetching Julia and 32-bit Windows on AppVeyor is due to WinRPM.

musm:
could we get a tag with these changes?

tkelman:
No, BinDeps master is not safe to tag from right now.

KristofferC:
What does this mean?

ararslan:
Tony explained in https://github.com/JuliaLang/METADATA.jl/pull/11896. There are a couple of PRs we need to get in first.

PR #349, opened 2018-01-14
pushfirst! not pushfront! 0.6 (closes #347)

### LINKED ISSUE #347: pushfront! not defined
Recent release that fix 0.7 warnings break builds 0.6 as `pushfront!` is not defined in 0.6

oxinabox:
to ensure fixed I added Gumbo (which uses this code path, though autotools) to the tests.

<s>prob would be better to use Compat for this, </s>

but for now (and to unbreak like dozens of packages ? (at least 4 to my knowledge))
this is what i have.

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -6,9 +6,6 @@ if !isdefined(Base, :pairs)
     pairs(x) = (a => b for (a, b) in x)
 end
 
-if !isdefined(base, :pushfront!)
-    pushfront!(xs, x) = unshift!(xs, x)
-end
 
 # A dependency provider, if successfully executed will satisfy the dependency
 abstract type DependencyProvider end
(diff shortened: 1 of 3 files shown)


src/dependencies.jl
```diff
@@ -6,6 +6,10 @@ if !isdefined(Base, :pairs)
     pairs(x) = (a => b for (a, b) in x)
 end
 
+if !isdefined(base, :pushfront!)
+    pushfront!(xs, x) = unshift!(xs, x)
```
fredrikekre: This is already in Compat v0.43 which is `REQUIRE`'d (#346)

oxinabox:
Compat has pushfirst.
So does Base.
https://github.com/JuliaLang/julia/blob/master/base/deprecated.jl#L2654

I don't think `pushfront` is  a thing.

fredrikekre:
Oh, seems like a mistake in #346 then. Presumably the fix is to change `pushfront!` -> `pushfirst!` instead?

PR #359, opened 2018-03-21
Use Compat.Sys.BINDIR so that BinDeps can work with system images

rofinn:


src/BinDeps.jl
```diff
@@ -122,9 +118,9 @@ if Sys.KERNEL == :FreeBSD
 end
 
 if Compat.Sys.iswindows()
-    const exe7z = joinpath(_HOME, "7z.exe")
```
tkelman: it's possible other packages may have been referring to this constant

rofinn: I really hope not since this would only existing on windows. I could undo this without causing problems for us, but I'm pretty sure creating system images on windows won't work correctly.

tkelman: > creating system images on windows won't work correctly

Why's that? What is this fixing anyway?

rofinn: The issue is that `JULIA_HOME` (or I guess now `Sys.BINDIR`) doesn't exist at system image build time, which will cause functions that depend on the compiled constant to fail and runtime. Since I don't use windows I'm not particularly bothered by leaving this in the windows if block, but this could cause problems for windows user who want to make system images.

tkelman:
it's not possible to preserve the compatibility and have both?

rofinn:
Hmmm, the only way we can keep the windows constant is by resetting it in an `__init__` function?

PR #368, opened 2018-08-04
add missing `catch`es [so it builds]

JeffreySarnoff:
two try blocks were missing catch
test/REQUiRE had HttpParser, I replaced that with URIParser

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.6
-URIParser
+HttpParser
 SHA
 Compat 0.62.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.6
-HttpParser
+URIParser
 SHA
 Compat 0.62.0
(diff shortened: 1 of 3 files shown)


staticfloat:
Why did you change it to URIParser? It looks to me like it should still be HTTPParser.

JeffreySarnoff:
(a) I thought URIParser exported the more up-to-date version of the `URI(..)` function
(b) |t seemed that the only use of HttpParser was for `URI`
(c) HttpParser does not build and was blocking this from working
Of those, (c) was true last night and (a), (b) were based on a cursory look by unfamiliar eyes
It makes no difference to me -- just need BinDeps to work so people who want to do so can use ArbNumerics after my talk.  Please make any changes that help -- I'm about to fly.

staticfloat:
Okay then, you will need to change the `using` within the `test/runtests.jl` file as well.

tkelman:
URIParser doesn't use BinDeps, HttpParser does. It's a test-only dependency and should be left alone.

JeffreySarnoff:
yes! it works with nightly

PR #376, opened 2018-08-10
Fix #375: no method matching similar(::Dict{Any,Any})

jayschwa:

PR #378, opened 2018-08-14
fix similar(env) deprecation in 0.7

stevengj:
`similar(env::AbstractDict)` is deprecated in favor of `empty(env)` in 0.7 (JuliaLang/julia#25224).  There is no Compat support for this, it seems, but this particular call site could be changed to use `copy`.

(I noticed this in updating CMakeWrappers.jl)


stevengj:
Duplicate of #376

PR #379, opened 2018-08-16
test on Julia 1.0 and fix some deprecations

rdeits:
x-ref: https://github.com/JuliaLang/METADATA.jl/pull/16899#issuecomment-413568001

Tests on Julia 1.0 are currently broken primarily because the BinDeps.jl tests try to build GSL.jl, but GSL hasn't been updated for Julia 1.0. Since it's better to know what the problem is, I've turned tests for v1.0 but marked them as allowed failures.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.6
 URIParser
 SHA
-Compat 0.65.1
+Compat 0.62
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.6
 URIParser
 SHA
-Compat 0.62.0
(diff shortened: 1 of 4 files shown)


REQUIRE
```diff
@@ -1,4 +1,4 @@
 julia 0.6
 URIParser
 SHA
-Compat 0.62.0
+Compat 0.65.1
```
tkelman: this can be left where it was, the 0.65.1 change was readme-only https://github.com/JuliaLang/Compat.jl/commit/2a822219a157b7fe0f45c6549fe4020c342ac29f

rdeits: done

PR #383, opened 2018-08-27
Only use /sbin/ldconfig if it exists.

rbvermaa:
On some platforms, /sbin/ldconfig does not exist, e.g. NixOS [0]. This patch skips the /sbin/ldconfig call if the path does not exist.

[0] https://nixos.org/nixos


rbvermaa:
Woult it be possible to get this merged? Is the package still maintained?

PR #386, opened 2018-09-28
fixes to fail to apt-get install

yukota:
apt-get require to type yes.
add -y option to skip confirm.


tkelman:
This is intentionally interactive, as apt-get can do surprising things. See #311 and others.

PR #389, opened 2018-11-23
Fix tests on Julia 1.0

### LINKED ISSUE #387: Outdated usage of `STDOUT` in BinDeps.debug
`BinDeps.debug()` looks like it's missing some v0.7 deprecations: 

```julia
julia> using BinDeps

julia> BinDeps.debug("LCMCore")
ERROR: UndefVarError: STDOUT not defined
Stacktrace:
 [1] debug(::String) at /home/rdeits/.julia/packages/BinDeps/ZEval/src/debug.jl:69
 [2] top-level scope at none:0
```

### LINKED ISSUE #388: Tests require HttpParser, which cannot be installed in Julia 1.0
HttpParser is version-capped (and in fact cannot be built on 1.0), so it can't be used for testing BinDeps.

rdeits:
Fixes #387 
Closes #379 
Fixes #388 by not building HttpParser anymore. If anyone can suggest a representative package to replace it with, then I'll add it. Otherwise I'll just try to pick something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ os:
 julia:
   - 0.6
   - 0.7
+  - 1.0
   - nightly
 notifications:
   email: false
 after_success:
-  - julia -e 'cd(Pkg.dir("BinDeps")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder()); Codecov.submit(Codecov.process_folder())';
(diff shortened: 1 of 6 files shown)


rdeits:
We could use https://github.com/JuliaRobotics/LCMCore.jl/blob/master/deps/build.jl as it always builds from source and has been pretty stable for a long time (and it actually works on Julia 1.0).

PR #393, opened 2018-12-06
[RFC] Remove reliance on absolute program paths

ninjin:
Breaks for Nix, NixOS, and other systems that rely on commands honouring `PATH`.


ninjin:
The CI errors appears to be unrelated.

PR #404, opened 2019-06-18
Use --yes flag for apt-get install

goretkin:
Addresses https://github.com/JuliaPackaging/BinDeps.jl/issues/403

### DIFF
--- a/src/dependencies.jl
+++ b/src/dependencies.jl
@@ -451,7 +451,7 @@ function generate_steps(dep::LibraryDependency,h::AptGet,opts)
     sudo = get(opts, :sudo, has_sudo[]) ? `sudo` : ``
     @build_steps begin
         println("Installing dependency $(h.package) via `$(sudoname(sudo))apt-get install $(h.package)`:")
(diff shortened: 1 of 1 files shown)


lobingera:
Is --yes side-effect free? (read as: should this be the default, or an option)

goretkin:
I don't know why this wouldn't be the default. Without it (or other option) apt-get is an interactive program, since it prompts users in some circumstances. I don't understand clearly which circumstances those are, to be honest. Maybe when there are many second-order dependencies.

tkelman:
see #311 etc, there have been many past discussions on this. when sudo is involved it's very dangerous to be doing system-wide things without the user paying attention

lobingera:
@tkelman So you recommend to have this only optional?

goretkin:
In my case, I just want to install packages in a docker image, and it asks to confirm that I want to install additional packages.

I don't understand the danger of `-y`, but if you claim it's dangerous, then we can be cautious.

The other option is to set the environment variable `DEBIAN_FRONTEND=noninteractive`, which I guess a user can already do.

PR #409, opened 2019-12-05
Bump version

musm:
cc @giordano

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "BinDeps"
 uuid = "9e28174c-4ba2-5203-b857-d8d62c4213ee"
-version = "1.0"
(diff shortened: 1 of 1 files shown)


giordano:
Maybe we want to go into 1.0 land?

musm:
Sounds like a good idea. I'm going to go ahead and do that since both you and @fredrikekre agree

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "BinDeps"
 uuid = "9e28174c-4ba2-5203-b857-d8d62c4213ee"
-version = "0.8.10"
+version = "1.0"
```
fredrikekre:
```suggestion
version = "1.0.0"
```

